PR #24, opened 2021-01-31
Adds some additional fixes/support for Lua

gmjosack:
* Provides a table to access entity names. table is named `ENT_TYPE`
  with entries named such as `ITEM_BOMB`.
* Pulled in patch from iojon to only re-execute script if it was updated
  in the UI
* Added new function called `setinterval` which takes a callback and
  milliseconds for how frequently to execute the callback.


iojon (approved):
Looks good to me!

PR #26, opened 2021-01-31
added spawn_door

C-ffeeStain:
Don't know why there are so many new files but feel free to remove any of them that are not needed. I've made changes mostly to ui.cpp. There is now a spawn_door function in the Lua VM that does the same thing as the one in Overlunky.

### DIFF
--- a/src/injected/ui.cpp
+++ b/src/injected/ui.cpp
@@ -2800,12 +2800,11 @@ PresentPtr &vtable_find(T *obj, int index) {
     return *reinterpret_cast<PresentPtr *>(&ptr[0][index]);
 }
 
-void spawn_door_lua(float x, float y, uint8_t l, uint8_t w, uint8_t f, uint8_t t)
-{
-    spawn_door(x, y, l, w, 1, t + 1);
-}
 
 void init_script() {
+    // lua.new_usertype<StateMemory>("StateMemory", "shoppie_aggro",
+    //                               &StateMemory::shoppie_aggro);
+    // lua["state"] = g_state;
     lua.open_libraries(sol::lib::math);
     lua.set_function("beep", [] {
         update_players();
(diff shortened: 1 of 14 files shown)


src/injected/ui.cpp
```diff
@@ -2800,29 +2800,58 @@ PresentPtr &vtable_find(T *obj, int index) {
...
+                              &StateMemory::merchant_aggro);
+    lua["state"] = g_state;
     lua.create_named_table("ENT_TYPE");
-    for (int i = 1; i < g_items.size(); i++) {
+    for (int i = 1; i < g_items.size(); i++)
+    {
         auto name = g_items[i].name.substr(9, g_items[i].name.size());
         lua["ENT_TYPE"][name] = g_items[i].id;
     }
+    lua.create_named_table("THEME");
```
gmjosack, line 2837: Since this is a static list you can use a read-only enum like

```lua

lua.new_enum("THEME",
    "DWELLING", 0,
    "JUNGLE", 1,
    <etc>.
);
```

src/injected/ui.cpp
```diff
@@ -2800,29 +2800,58 @@ PresentPtr &vtable_find(T *obj, int index) {
...
-            std::chrono::duration<unsigned int, std::milli>(ms),
+            std::chrono::duration(ms),
             std::chrono::system_clock::time_point::min(),
         };
         g_luaCallbacks.push_back(luaCb);
     });
-    lua["spawn_entity"] = spawn_entity;
 
+    lua["spawn_entity"] = spawn_entity;
+    lua["spawn_door"] = spawn_door_lua;
```
gmjosack, line 2825: where is `spawn_door_lua` defined?

gmjosack (changes requested):

PR #27, opened 2021-02-01
Added spawn_door to Lua VM

C-ffeeStain:
I have added the function spawn_door and an enum for it in the Lua VM.

### DIFF
--- a/src/injected/ui.cpp
+++ b/src/injected/ui.cpp
@@ -2799,33 +2799,63 @@ PresentPtr &vtable_find(T *obj, int index) {
     return *reinterpret_cast<PresentPtr *>(&ptr[0][index]);
 }
 
-void init_script() {
+void spawn_door_lua(float x, float y, uint8_t l, uint8_t w, uint8_t f, uint8_t t)
+{
+    spawn_door(x, y, l, w, 1, t + 1);
+}
+
+void init_script()
+{
     g_state = (struct StateMemory *)get_state_ptr();
     g_state_addr = reinterpret_cast<uintptr_t>(g_state);
-
-    lua.open_libraries(sol::lib::math);
+    lua.open_libraries(sol::lib::base, sol::lib::math);
     lua.set_function("beep", [] {
         update_players();
-        for (auto player : g_players) {
+        for (auto player : g_players)
+        {
             render_hitbox(player, false, ImColor(255, 0, 255, 200));
         }
     });
     lua.set_function("setinterval", [](sol::function cb, int ms) {
-        auto luaCb = LuaIntervalCallback{
+        auto luaCb = LuaIntervalCallback {
             cb,
             std::chrono::duration<unsigned int, std::milli>(ms),
             std::chrono::system_clock::time_point::min(),
         };
         g_luaCallbacks.push_back(luaCb);
     });
-    lua["spawn_entity"] = spawn_entity;
-    lua["shopAggro"] = &g_state->shoppie_aggro;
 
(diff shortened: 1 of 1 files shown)


src/injected/ui.cpp
```diff
@@ -2799,33 +2799,63 @@ PresentPtr &vtable_find(T *obj, int index) {
...
+                              &StateMemory::merchant_aggro);
+    lua["state"] = g_state;
     lua.create_named_table("ENT_TYPE");
-    for (int i = 1; i < g_items.size(); i++) {
+    for (int i = 1; i < g_items.size(); i++)
+    {
         auto name = g_items[i].name.substr(9, g_items[i].name.size());
         lua["ENT_TYPE"][name] = g_items[i].id;
     }
+    lua.create_named_table("THEME");
```
gmjosack, line 2841: I thought you were going to use an enum for this?

src/injected/ui.cpp
```diff
@@ -2799,33 +2799,63 @@ PresentPtr &vtable_find(T *obj, int index) {
...
+    spawn_door(x, y, l, w, 1, t + 1);
+}
+
+void init_script()
+{
     g_state = (struct StateMemory *)get_state_ptr();
     g_state_addr = reinterpret_cast<uintptr_t>(g_state);
-
-    lua.open_libraries(sol::lib::math);
+    lua.open_libraries(sol::lib::base, sol::lib::math);
```
gmjosack, line 2811: I don't see anything in the commit or comments on why `base` is being added here and I'm having a hard time finding a comprehensive doc on what functions it includes.

src/injected/ui.cpp
```diff
@@ -2799,33 +2799,63 @@ PresentPtr &vtable_find(T *obj, int index) {
     return *reinterpret_cast<PresentPtr *>(&ptr[0][index]);
 }
 
-void init_script() {
+void spawn_door_lua(float x, float y, uint8_t l, uint8_t w, uint8_t f, uint8_t t)
+{
+    spawn_door(x, y, l, w, 1, t + 1);
```
gmjosack, line 2804: Why even take `f` in the lua wrapper if you're just going to pass one? And why do you add 1 to t? Is there some reason the lua api should behave differently from the game api? If there's a good reason for it, that's something worth commenting because otherwise it's just confusing.

gmjosack (changes requested):

src/injected/ui.cpp
gmjosack: I believe http://www.lua.org/manual/5.3/manual.html#6.1 covers the scope of what gets pulled in for base. I'm not sure we want to make all of that available just to provide `print` personally

src/injected/ui.cpp
gmjosack: I believe http://www.lua.org/manual/5.3/manual.html#6.1 covers the scope of what gets pulled in for base. I'm not sure we want to make all of that available just to provide print personally

src/injected/ui.cpp
gmjosack: Talking in Discord it seems it's because the all themes are +1 from the enum. Seems to make more sense to me for the enum be updated to start at 1 instead of +1 everywhere.

src/injected/ui.cpp
C-ffeeStain: was going to, adding now.

src/injected/ui.cpp
C-ffeeStain: removed base from being included here

PR #32, opened 2021-03-10
Refactor of dx11 and window hooks

Malacath-92:
Raw input breaks with this PR, hence it's WIP. But I don't know how that is supposed to work so it's difficult to debug.

Other than that it seems to fix alt-tab and changing to fullscreen. Plus it moves the imgui and dx11 hooks to game_api so that Playlunky can eventually profit from it.

### DIFF
--- a/src/game_api/CMakeLists.txt
+++ b/src/game_api/CMakeLists.txt
@@ -15,6 +15,8 @@ add_library(game_api STATIC
         memory.hpp
         thread_utils.cpp
         thread_utils.hpp
-        movable.hpp)
+        movable.hpp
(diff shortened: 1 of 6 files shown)


iojon (approved):

Malacath-92:
I played around with it for ~30min and couldn't really find any issues, but I am not familiar with all the OL features or how the GUI should look so I leave that judgement up to y'all.

Malacath-92:
Raw input is only hooked if the game actually hooks it btw, so nothing should happen in case one plays with a game pad, but I don't have any right now to test.

PR #33, opened 2021-03-11
Move script.hpp/.cpp to game_api

Malacath-92:
Only moves files and hides sol away so game_api doesn't have to pass that dep down. Just wanted to get this step out the way and from there move more things from ui.cpp into script.cpp because this first step is more likely to cause headache merge conflicts.


iojon (approved):
lgtm

Issue #28, opened 2021-02-04
Lua API features and goals


Dregu:
Feel free to dump your ideas and required functions for the Lua scripting API here.

sjtower:
Thank you for all your work, Dregu! 

I don't know anything about Lua scripts, what kinds of things could we use them for in Spelunky2? 
Just off the bat, it would be cool if we could load in enemies and traps that are not in the level editor. This would require integrating more tightly with the editor.

d11z:
I think the Lua API should allow access to primitive drawing using imgui (line, text, etc). Or maybe even simpler, just a boolean to toggle showing an entity's hitbox.

0novo:
An enum for search flags would be a great addition

Mr-Auto:
Change so the **entity_has_item_type** returns the entity/uid of the item instead of bool, 
I'm pretty sure you can only have one item of a certain type, and nil if entity doesn't have it. Two in one function. Or make new function for it

Can we also get event's **on_entity_spawn** with uid as argument (ideally run before the item is drawn on screen), **on_reset**, **on_entity_kill** or **before_entity_kill** with uid, actually why events never have arguments? can they have them even? like on_level giving you a level number, on_win giving info what win it is etc.

paul-lindberg:
Currently, spawning `LIQUID_*` entities with `spawn_entity()` results in a crash. A fix for this or a workaround would be appreciated.
 
 Another related topic; having increased liquid pools for areas in which liquid is problematic would be useful.
 For instance, currently on olmec, any liquid other than `LIQUID_STAGNANT_LAVA` won't work.

0novo:
Some things that would be nice to have:

1. Expand BUTTON enum to include movement (e.g. left = 0x100, right = 0x200, up, down), also maybe change the numbers to hex on the script doc (currently it goes 1, 2, 4, 8, 16, 32 instead of 0x1, 0x2, 0x4, 0x8, 0x10, 0x20)
2. PLAYER_STATE (CHAR_STATE?) enum (e.g. number = STANDING, otherNumber = HANGING, etc.)
3. AI_STATE enum (e.g. number = IDLING, etc.) <-- not sure if this one would work, i'm assuming that every entity uses the same values for their ai_states

hopefully shouldn't be too hard to implement :p

Mr-Auto:
[...]

PR #34, opened 2021-03-14
Change container class' inside variable to be 64-bit for the Lua API

0novo:
Hopefully lets an entity's special attributes be more manipulable via the Lua API when it's more than 32 bits. 
This would let us manipulate arrows to make them poisoned and anything else with more than 32 bits used (e.g. bombs, crush traps) via the Lua API.
I've tested a build with this change and from what I've seen nothing was wrong.
I understand that the scripting API will change. This is just a quick and easy fix for a simple problem.

### DIFF
--- a/src/game_api/movable.hpp
+++ b/src/game_api/movable.hpp
@@ -115,7 +115,7 @@ class Player : public Monster
 class Container : public Movable
 {
   public:
-    int32_t inside;
+    int64_t inside;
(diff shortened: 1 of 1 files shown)


Dregu:
Nice try but not like this. I'll try to make real entity types for the guns, crushblocks and what have you soon. Feel free to document those in #28 if you know something we don't.

PR #35, opened 2021-03-15
Changes for playlunky compatibility

Malacath-92:
- renamed game_api to spel2_api (because less generic name)
- added spel2.dll (C wrapper around spel2_api)
- renamed Script to SpelunkyScript (because less generic name)
- split script draw and update
- added cmake interface to avoid building overlunky (by default builds overlunky and not spel2.dll)


Malacath-92:
Uhhh, I don't know what happened with those commits from before my changes, I rebased so they shouldn't look like this...

Malacath-92:
Also @iojon what's the reason for not using glob patterns for the projects in cmake? Using CONFIGURE_DEPENDS the project will regenerate correctly even when adding new files.

iojon:
Ah I removed glob pattern in dcfe7ca before because it required cmake to run configure step which checked if globbed files are changed. However, since at least one project are now using glob, I think it's fine to use it.

Malacath-92:
Which project is using glob? Some of the submodules?

iojon:
[injector](https://github.com/spelunky-fyi/overlunky/blob/main/src/injector/CMakeLists.txt#L1) seems to be using it.

Malacath-92:
Was that maybe just an oversight when removing it from the other projects?

iojon:
c96ef14 added glob to injector, which was 6 hours after glob removal. So I guess so...
I previously believed that it was making the build much slower, but soon I found that it wasn't.
Either way is fine with me.

Malacath-92:
Oops, sorry about that. I thought that it would only make configuring slower, I usually disable the zero check on my projects so it doesn't affect the build but you're right that with the zero check the build will actually be slower.

iojon (approved):
Oh, I didn't know about the zero check feature.. 😅 lgtm

Malacath-92:
I think I need to do some small changes before pl support is all there. So if you can hold on for a day before merging it'd be good.
Also do we feel ready to split the api stuff into a separate lib? Or should we keep it with OL for now since that's what's driving the development mainly.

Malacath-92:
Now this is all that PL needs from how it looks right now. Ideally we would split up the api stuff into its own library and have PL and OL depend on that. Idk what's the best place to have that discussion tho.

PR #43, opened 2021-03-29
implement movable::is_poisoned() and movable::poison() to control …

zappatic:
…poisoned status

### DIFF
--- a/src/game_api/entity.cpp
+++ b/src/game_api/entity.cpp
@@ -225,6 +225,23 @@ void Mount::tame(bool value)
     flags = flags | 0x20000;
 }
 
+void Movable::poison(bool p)
+{
+    if ( p )
+    {
+        poison_tick_timer = 0x0708; // = 1800 = 30 seconds 
+    }
+    else
+    {
+        poison_tick_timer = 0xFFFF; // disables poison
+    }
+}
+
+bool Movable::is_poisoned()
+{
+    return (poison_tick_timer != 0xFFFF);
+}
+
 void Entity::destroy()
 {
     // TODO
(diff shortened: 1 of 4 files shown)


src/game_api/entity.cpp
```diff
@@ -225,6 +225,23 @@ void Mount::tame(bool value)
     flags = flags | 0x20000;
 }
 
+void Movable::poison(bool p)
+{
+    if ( p )
+    {
+        poison_tick_timer = 0x0708; // = 1800 = 30 seconds
```
gmjosack, line 232: If this can be passed any interval it might be nice to expose this to the api

zappatic: You could, but I think once the timer reaches zero to deal you a tick of damage, it will reset to 30 seconds again, that's why I hardcoded 0x0708. Of course, we could always overwrite the hardcoded 0x0708 in the reset function perhaps... I'll take a look at this tomorrow.

gmjosack: ah okay so it's the number of ticks left not the interval. Still think it could be cool to expose a way to set it since from lua you could manage your own custom timers and even just track and reset it once the game sets it back.

zappatic: Yes, basically poisoned status is that timer not being 0xFFFF. It counts down every frame and when it reaches 0 then it deals a point of damage, and resets it to 30 seconds. It shouldn't be difficult to change the instruction that resets to 0x0708, as we just have to replace that value with anything below 0xFFFF, and it will keep resetting to the user requested value. Don't merge this yet.

PR #46, opened 2021-04-01
Dump

Malacath-92:
Sorry @iojon for hijacking your branch. I think this is worth merging into main just to ensure it will keep working with future changes and can easily be extended if needed. The alternative is to keep it in its own branch and constantly merge main to make sure it doesn't break.

Adds a info_dump.dll that dumps (for now)
- entities.json containing the known info about entities as a map from ent name to data
- textures.json all relevant(?) data of textures that the game loads as a map from id to data 

Adding @gmjosack to the review in case you have some input on any other data for textures that could be relevant


gmjosack (approved):

Malacath-92:
Forgot to attach the current output. Should ofc be .json and not .txt  🤷‍♀️ 
[textures.txt](https://github.com/spelunky-fyi/overlunky/files/6244737/textures.txt)
[entities.txt](https://github.com/spelunky-fyi/overlunky/files/6244738/entities.txt)

iojon (approved):
lgtm

PR #48, opened 2021-04-02
Added Olmec details and timings

zappatic:


Dregu:
I'm pretty sure some of this doesn't work when there's multiple olmecs (mods can do that) but we can fix it later...

PR #56, opened 2021-04-16
ongoing changes

zappatic:


src/game_api/rpc.cpp
```diff
@@ -943,3 +943,91 @@ void pick_up(uint32_t who, uint32_t what)
...
+    {
+        auto state = State::get();
+        auto memory = Memory::get();
+        auto exe = memory.exe();
+
+        std::string pattern_phase1 = "\xF3\x0F\x10\x15\x07\x3D\x36\x00"s;
+        std::string pattern_phase2 = "\xF3\x0F\x10\x0D\xCD\x38\x36\x00"s;
+
+        // first look up these patterns so we are in the correct function
+        auto offset1 = find_inst(exe, pattern_phase1, memory.after_bundle);
```
Malacath-92, line 968: Should these be stored somewhere statically?

src/game_api/rpc.cpp
```diff
@@ -943,3 +943,91 @@ void pick_up(uint32_t who, uint32_t what)
...
+    static size_t normal_offset = 0;
+    static size_t cursed_offset = 0;
+    if (normal_offset == 0)
+    {
+        auto state = State::get();
+        auto memory = Memory::get();
+        auto exe = memory.exe();
+
+        std::string pattern = "\xBA\x28\x23\x00\x00\x41\xB8\x30\x2A\x00\x00"s;
+        auto offset = find_inst(exe, pattern, memory.after_bundle);
```
Malacath-92, line 1027: Same with this

src/game_api/state.hpp
```diff
@@ -53,7 +54,7 @@ struct StateMemory
     int32_t i70;
     uint8_t theme;
     uint8_t theme_next;
-    uint8_t b72;
+    uint8_t win_state; // 0 = no win 1 = tiamat win 2 = hundun win 3 = CO win; set this and next doorway leads to victory scene
```
Malacath-92, line 57: Does it make sense to add an enum for the Lua API for this?

src/game_api/rpc.cpp
zappatic: These are intermediary/temporary offsets, and serve only to calculate phase1_offset and phase2_offset, which are static and permanent. The whole pattern detection block is only executed once (if the pattern is found of course) and there is no further use for offset1 and 2, so they don't need to be stored permanently.

src/game_api/state.hpp
zappatic: Good point, I'll add it

src/game_api/rpc.cpp
Malacath-92: Oh yes, I didn't read the start of this block properly. It's not looking how I'm used to from OL so I just missed it.

docs/script-api.md
```diff
@@ -731,6 +731,7 @@ Derived from [`Entity`](#entity) [`Movable`](#movable) [`Monster`](#monster) [`C
 - [`saved_hamsters`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=saved_hamsters) &StateMemory::saved_hamsters
 - [`win_state`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=win_state) &StateMemory::win_state
 - [`illumination`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=illumination) &StateMemory::illumination
+- [`money_total`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=money_total) &StateMemory::money_total
```
jeremypepper, line 734: I believe that this file is generated from generate.py, I think we will want to edit that to update this script-api.

zappatic: This file is indeed auto-generated. Can you clarify what exactly needs to be edited/updated?

jeremypepper: oh you know what I think this part will get updated, please ignore me :)

src/game_api/state.hpp
```diff
@@ -72,7 +72,7 @@ struct StateMemory
     uint32_t time_last_level;
     uint32_t time_level;
     int32_t ia00;
-    int32_t ia04;
+    uint32_t money_total;
```
jeremypepper, line 75: So it appears this won't be the total money that appears in the top right of the screen but the money from the previous levels. I wonder if it would be possible to name this something else or provide additional documentation. Maybe would be less confusing if this was named `money_last` to note that is the money total at the end of the last level.

zappatic: Sure, I've named it money_last_levels, to make it slightly more descriptive

PR #57, opened 2021-04-19
Script options overhaul

Malacath-92:
Rewrites the options code to be a bit more idiomatic (because I saw that TODO in code)
Adds an optional long_desc to the options which is rendered below the option using `ImGui::TextWrapped` so people stop complaining that they can't write their essays in option descriptions.


src/game_api/script.cpp
```diff
@@ -999,6 +1065,14 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         &Inventory::kills_level,
         "kills_total",
         &Inventory::kills_total);
+    lua.new_usertype<Animation>(
```
Malacath-92, line 1068: Oh, this was not meant to be part of this PR... I can revert it if you want.

src/game_api/script.cpp
```diff
@@ -1028,7 +1102,9 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         "life",
         &EntityDB::life,
         "blood_content",
-        &EntityDB::blood_content);
+        &EntityDB::blood_content,
+        "animations",
+        &EntityDB::animations);
```
Malacath-92, line 1107: Same here ofc

PR #60, opened 2021-04-23
updates docs to clarify overlunky.ini location

jeremypepper:
Updates script-api.md to make the location of the overlunky.ini file more clear.

PR #61, opened 2021-04-23
Temp playlunky branch

Malacath-92:
Adds:
- Entity::overlaps_with(Entity)
- Entity::overlaps_with(left, bottom, right, top) // manual collision rect
- get_entities_overlapping(type, mask, left, bottom, right, top, layer)
- ON.GAMEFRAME which runs also in camp, transition and death
- MASK.ROPE
- search_flags.json (probably useless, but it was done when I realized that)
- docs update from #60 

Fixes:
- crash when running daily
- typo
- memory leak when reloading scripts


gmjosack (approved):

src/game_api/entity.hpp
```diff
@@ -211,6 +211,31 @@ class Entity
         return topmost;
     }
 
+    bool overlaps_with(float rect_left, float rect_bottom, float rect_right, float rect_top)
+    {
+        const auto [posx, posy] = position();
+        const float left = posx - hitboxx + offsetx;
+        const float right = posx + hitboxy + offsetx;
```
zappatic, line 218: Should this not be hitboxx ?

src/game_api/entity.hpp
```diff
@@ -211,6 +211,31 @@ class Entity
...
+            && rect_left < right
+            && bottom < rect_top
+            && rect_bottom < top;
+    }
+
+    bool overlaps_with(Entity* other)
+    {
+        const auto [other_posx, other_posy] = other->position();
+        const float other_left = other_posx - other->hitboxx + other->offsetx;
+        const float other_right = other_posx + other->hitboxy + other->offsetx;
```
zappatic, line 232: Same here? hitboxx instead of y

src/info_dump/main.cpp
```diff
@@ -166,5 +166,32 @@ extern "C" __declspec(dllexport) void run(DWORD pid)
         textures_file.write(dump.data(), dump.size());
     }
 
+    if (std::ofstream search_flags_file = std::ofstream("game_data/search_flags.json"))
+    {
+        float_json search_flags(float_json::object());
+        for (size_t i = 31; i < 32; i--)
+        {
+            auto items = list_entities();
+            std::sort(items.begin(), items.end(), [](EntityItem &a, EntityItem &b) -> bool { return a.id < b.id; });
```
zappatic, line 175: [pedantic] listing entities and sorting them could be placed before the for loop

zappatic (approved):

src/game_api/entity.hpp
Malacath-92: Yes it should   :)
Good catch

src/game_api/entity.hpp
Malacath-92: All my tests worked because many entities have similar x and y  😅

src/info_dump/main.cpp
Malacath-92: Uhh yes, could even be done fore all the json stuff. Will fix.

PR #62, opened 2021-04-24
Update script-api.md

Mr-Auto:
tiding up functions with alternative names


gmjosack:
This file is auto-generated from https://github.com/spelunky-fyi/overlunky/blob/main/docs/generate.py so any changes have to be made to the generator and not to the file directly

PR #63, opened 2021-04-25
Save/Load Mod API

C-ffeeStain:
I have created a fairly simple API for mods to save and load data from a file called save.dat. The users can save to the file using `save_data("data")` and load its contents using `load_data()`.

### DIFF
--- a/src/game_api/script.cpp
+++ b/src/game_api/script.cpp
@@ -867,11 +867,11 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         return sol::nullopt;
     };
 
-    /// Loads save.dat from the script's pack. If save.dat does not exist, you will get an empty string.
+    /// Loads save_data_{meta.name}.dat from the script's pack. If save_data_{meta.name}.dat does not exist, you will get an empty string.
     lua["load_data"] = [this]() -> std::string
     {
         std::ifstream datafile;
-        datafile.open(script_folder / "save.dat");
+        datafile.open(script_folder / ("save_data_" + meta.name + ".dat"));
         if (datafile.good()) 
         {
             std::string string;
@@ -891,11 +891,11 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         }
     };
 
-    /// Saves `data` to the save.dat file in the script's pack.
+    /// Saves `data` to save_data_{meta.name}.dat, which is in the script's pack. If save_data_{meta.name}.dat does not exist, it will be created.
     lua["save_data"] = [this](std::string data)
     {
         std::ofstream datafile;
-        datafile.open(script_folder / "save.dat");
+        datafile.open(script_folder / ("save_data_" + meta.name + ".dat"));
         datafile << data;
         datafile.close();
     };
--- a/src/game_api/script.cpp
+++ b/src/game_api/script.cpp
@@ -874,7 +874,7 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
     };
 
     /// Loads save.dat from the script's pack. If save.dat does not exist, you will get an empty string.
-    lua["load_data"] = [this](std::string path) -> std::string
+    lua["load_data"] = [this]() -> std::string
     {
         std::ifstream datafile;
         datafile.open(script_folder / "save.dat");
--- a/src/game_api/script.cpp
+++ b/src/game_api/script.cpp
@@ -873,7 +873,7 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         return sol::nullopt;
     };
 
-    /// Loads the `save.dat` file from the script's pack. If `save.dat` does not exist, you will get an empty string.
+    /// Loads save.dat from the script's pack. If save.dat does not exist, you will get an empty string.
     lua["load_data"] = [this](std::string path) -> std::string
     {
         std::ifstream datafile;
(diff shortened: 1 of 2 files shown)


src/game_api/script.cpp
```diff
@@ -168,6 +168,11 @@ ImVec2 screenify(ImVec2 pos)
     return screened;
 }
 
+bool exists(const std::string& filename) {
```
Malacath-92, line 171: Should probably use `std::filesystem::exists`

src/game_api/script.cpp
```diff
@@ -867,6 +872,39 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
...
+        datafile.open(script_folder / "save.dat");
+        if (datafile.good()) 
+        {
+            std::string string;
+
+            datafile.seekg(0, std::ios::end);   
+            string.reserve(datafile.tellg());
+            datafile.seekg(0, std::ios::beg);
+
+            string.assign((std::istreambuf_iterator<char>(datafile)),
```
Malacath-92, line 888: I've had some bad experiences with this API for reading whole files in. Just to keep in mind that if there are ever issues mentioned we might want to move over to the C API (the only C API that I personally prefer to the C++ API 😄)
No need to act on this now.

Malacath-92 (changes requested):

src/game_api/script.cpp
C-ffeeStain: oops! I forgot to remove that function! My apologies.

src/game_api/script.cpp
C-ffeeStain: Alright.

gmjosack:
I'm not sure what other game scripting libraries do for save apis but i'm slightly concerned about people saving every frame, or even worse every guiframe, and burning out write ops on my SSD. Some thoughts there would be to limit saves to certain callbacks like RESET, LEVEL, etc. Specifically limiting it from the FRAME events. If that idea isn't liked, alternatively we could do some kind of write coalescing, buffering the save data and flushing at some interval, or debouncing the write request, which some max number of debounces to ensure flushing eventually happens.

Another thing, this seems to prevent multiple mods to be used with OL, is that an intentional part of the design?

Lastly I'd like to think about how this will work once we add mod upgrades to Modlunky. Do we want to store this outside of the pack's directory when running via Playlunky? The thought there being once we add mod upgrading we'll need to purge the pack to ensure we're not leaving files that were removed from the mod but if state is kept in the pack folder itself then we have to ferry that state between upgrades. Alternatively one benefit of keeping this state in the pack is people can ship defaults with their pack. So i guess there's an open question of whether this is a desirable or undesirable feature having state stored inside the mod distribution.

Malacath-92:
> I'm not sure [...] eventually happens.
Very valid points, but I think that's perhaps a future improvement we can make?

> Another thing, this seems to prevent multiple mods to be used with OL, is that an intentional part of the design?
Good point, I hadn't thought about this yet, but I was gonna test this with OL later which might have made me notice  😅  OL is probably not a great place to distribute mods with, but either way. An easy fix would be to just rename the file to `save_data_{meta.name}.dat`. What I had thought initially (but didn't wanna bring up because it would complicate the work significantly) was to provide an interface that determines how stuff is saved. So OL could decide to save in another place wiht another name than PL does.

> Lastly I'd like to think about how this will work once we add mod upgrades to Modlunky.
Sort relevant what I wrote on top. Personally, I don't have strong feelings either way. Both have upsides and downsides you describe. Though imho shipping with "default" save game makes relatively little sense because you can just init state in the Lua file anyways. And you'll kinda have to do that in order to get an initial save anyways.

gmjosack:
> Very valid points, but I think that's perhaps a future improvement we can make?

For sure it's a future improvement but having seen what people do I don't think it's much of a stretch that people will immediately start using this to save data ON.FRAME because they just wanna also update their save data and don't think about it. I'm not terribly trusting of mods not to just try to write files 60 times per second which makes me feel like I'd have to audit mods for whether they use this API before using them moving forward. Maybe that's just me being overly cautious against an API that can cause writes to my filesystem

> So OL could decide to save in another place wiht another name than PL does.

heh, I had the same Idea and honestly I think it's something we'll probably want but I agree it's probably low priority for something like this. This should work as is with OL as long as only one script at a time is active that uses the save API. We haven't even really seen what people wanna do with this API yet so it's hard to say if people will want to run multiple in OL so maybe it's okay to defer.

> Though imho shipping with "default" save game makes relatively little sense because you can just init state in the Lua file anyways. And you'll kinda have to do that in order to get an initial save anyways.

I agree but it's a use case I've seen mentioned in the modding channel so figured I'd mention it. If we agree that shipping with default save data makes sense then ideally we would store the save data outside of the pack directory as I generally think state should be stored separately from the distribution itself but that would require different implementations between PL and OL.

Malacath-92:
@C-ffeeStain are you comfortable with adding some of the changes that @gmjosack suggested? I can get in on some of it myself if you don't want to. In particular the part of adding an API that PL and OL can customize might be a bit more complicated.

C-ffeeStain:
@Malacath-92 
you can help. I have absolutely no clue how to do most of what gary has suggested.

Malacath-92:
@C-ffeeStain @gmjosack I have submitted my suggested changes [to here](https://github.com/spelunky-fyi/overlunky/commit/3a88b18497d146759e5e2a7f10fe6c8857f1e786 because I couldn't manage to submit them to this PR (probably too late for me) We can talk about it here or wherever when we're all awake again.

Malacath-92:
@C-ffeeStain All these changes are now in main, by my own error without having merged the PR. We should probably close this PR, because your commits have anyways been merged already. Sorry again about my derping this up and thanks again for taking the initiative on this feature.

PR #65, opened 2021-04-27
Adds money_shop_total to script state

jeremypepper:
Adds money_shop_total which is the total $ spent in shops. This is expected
to be a negative value.


gmjosack:
While we don't exactly promise backwards compatibility i'm a little hesitant to expose an API that be definition will break if people figure out what the fields are for. I'd not want people to start distributing mods that use unknown fields.

jeremypepper:
Updated to fix merge conflicts and name variables closer to what zappatic did in their change

jeremypepper:
Updating original comment to reflect the current state of the PR.

jeremypepper:
Rebased and squished my changes down to a single commit. I tried to get the script-api.md to show up as a file but I was unsuccessful. Github's "rich diff" option in the top right corner is pretty good at pointing out the difference in the rendered markdown though.

PR #67, opened 2021-05-01
Load list of properties and events

Malacath-92:
Adds the following features to the Lua API
- `VANILLA_SOUND` enum that is filled with all the Fmod event names that the game defines
- `VANILLA_SOUND_PARAM` enum that is filled with all the parameter names that the game defines
- `VANILLA_CALLBACK_TYPE` enum that is filled some select types of callbacks that Fmod allows to hook
- `get_sound` that gets an already loaded sound based on a path or a vanilla sound based on `VANILLA_SOUND`
- `set/clear_vanilla_sound_callback` to allow hooking into vanilla sounds, allows for example to know when the player whips and then cancel that sound and play another one instead or play a custom sound when some vanilla sound is over
- `CustomSound::get_parameters()` that returns a list of parameters available for the sound, iff it is a vanilla sound. This can be used for debugging mainly.
- `PlayingSound::get_parameters()` is the same
- `PlayingSound::get_parameter(VANILLA_SOUND_PARAM)` gets the float value of the given parameter, or nil if the sound does not have that parameter. Always returns nil for custom sounds.
- `PlayingSound::set_parameter(VANILLA_SOUND_PARAM, float)` sets the parameter, returns false when the sound does not have that parameter or if it's a custom sound.
- Introduces a "global" lock for each Lua VM in order to keep the Fmod callbacks thread safe (I called it `gil`  😅) If no such Fmod callbacks are used it has virtually no overhead, and even when they are I couldn't measure any overhead.
I'm happy to hear any suggestions for better naming than "vanilla sound" (and "gil", but at least that's not user facing)


gmjosack:
I know this is still a WIP but noticed a lot of mixing between tabs and spaces for indentations :grimacing:

Malacath-92:
@gmjosack too attentive  :P
I'll run clang-format eventually, but I can't be arsed doing it all the time. Sorry for your trauma  🙏

Malacath-92:
@gmjosack and @zappatic
I know this is probably one of those "too big" PRs, but I am still interested in your opinion mainly on the API usability and naming. If you wanna look at the implementation feel free, I like any feedback  :)
And don't stress reviewing this, we're not in a hurry

PR #72, opened 2021-05-13
Adds reload_enabled_scripts shortcut

jeremypepper:
Shortcut is by default same as playlunky (ctrl +f5)


Malacath-92 (changes requested):

src/injected/ui.cpp
```diff
@@ -120,6 +120,7 @@ std::map<std::string, int> keys{
     {"mouse_clone", 0x505},
     {"mouse_destroy", 0x405},
     {"mouse_destroy_unsafe", 0x605},
+    {"reload_enabled_scripts", 0x173} // ctrl + f4 same as playlunky
```
gmjosack, line 123: Doesn't playlunky use ctrl + f5 to reload? I though ctrl + f4 only showed the menu

https://github.com/spelunky-fyi/Playlunky/blob/d98f09b1bdd680437248913524dd74d7dbc1c7d2/source/playlunky/mod/mod_manager.cpp#L574

Malacath-92: Yes, you are right. That is my mistake, I believe I said that on Discord.

src/injected/ui.cpp
```diff
@@ -3946,3 +3951,14 @@ void init_ui()
...
+void reload_enabled_scripts()
+{
+    for (auto& [script_name, script] : g_scripts)
+    {
+        if (script->is_enabled())
+        {
+            load_script(script->get_file(), true);
+        }
+    }
+}
```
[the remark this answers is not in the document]
jeremypepper: done. can we configure clang-format to do this?

src/injected/ui.cpp
jeremypepper: fixed

Malacath-92 (approved):

src/injected/ui.cpp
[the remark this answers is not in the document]
Malacath-92: A bit late response, but clang-format does not have an option for new-lines at eof. Idk why, but it's a contentious topic.

PR #75, opened 2021-05-18
Script: Adds read_stolen_input()

jeremypepper:
Adds a script function to Read input that has been previously stolen
with steal_input.

I don't understand why the input->orig_input->next starts off with a
value of 7712 but I've ignored this value for now.

### DIFF
--- a/docs/script-api.md
+++ b/docs/script-api.md
@@ -481,6 +481,11 @@ Send input
 ### [`read_input`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=read_input)
 #### Params: `int uid`
 Read input
+### [`read_stolen_input`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=read_stolen_input)
+#### Params: `int uid`
+<summary>
+Read input that has been previously stolen with steal_input
+</summary>
 ### [`window`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=window)
 #### Params: `string title, float x, float y, float w, float h, bool movable, function callback`
 #### Returns: `bool` (false if the window was closed from the X)
--- a/src/game_api/script.cpp
+++ b/src/game_api/script.cpp
@@ -1037,6 +1037,29 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         }
         return (uint16_t)0;
     };
+    /// <summary>
+    /// Read input that has been previously stolen with steal_input
+    /// </summary>
+    lua["read_stolen_input"] = [this](int uid)
+    {
+        Player* player = get_entity_ptr(uid)->as<Player>();
+        if (player == nullptr)
+            return (uint16_t)0;
+        ScriptInput* readinput = reinterpret_cast<ScriptInput*>(player->input_ptr);
+        if (!IsBadReadPtr(readinput, 20))
+        {
+            readinput = reinterpret_cast<ScriptInput*>(readinput->orig_input);
+            if (!IsBadReadPtr(readinput, 20))
+            {
+                // idk what 7712 is but it seems like it's an uninitialized default value that we should ignore
+                if (readinput->next != 7712)
+                {
+                    return readinput->next;
+                }
+            }
+        }
+        return (uint16_t)0;
+    };
 
     /// Returns: `bool` (false if the window was closed from the X)
     /// Create a new widget window. Put all win_ widgets inside the callback function. The window functions are just wrappers for the


docs/script-api.md
```diff
@@ -481,6 +481,11 @@ Send input
 ### [`read_input`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=read_input)
 #### Params: `int uid`
 Read input
+### [`read_stolen_input`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=read_stolen_input)
+#### Params: `int uid`
+<summary>
```
Malacath-92, line 486: What's the deal with these? I've seen one before and it seems to [break the markdown](https://github.com/jeremypepper/overlunky/blob/main/docs/script-api.md#read_stolen_input)

src/game_api/script.cpp
```diff
@@ -1037,6 +1037,29 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
...
+        if (player == nullptr)
+            return (uint16_t)0;
+        ScriptInput* readinput = reinterpret_cast<ScriptInput*>(player->input_ptr);
+        if (!IsBadReadPtr(readinput, 20))
+        {
+            readinput = reinterpret_cast<ScriptInput*>(readinput->orig_input);
+            if (!IsBadReadPtr(readinput, 20))
+            {
+                // idk what 7712 is but it seems like it's an uninitialized default value that we should ignore
+                if (readinput->next != 7712)
```
Malacath-92, line 1055: Is this value perhaps just the default buttons? Say if you were to not have run by default, does this value look different?

jeremypepper: If I have no buttons pressed the value is 0. I wonder if this is an and of all the buttons?

docs/script-api.md
jeremypepper: No idea I just ran the python script :(

src/game_api/script.cpp
jeremypepper: looks like it was the <summary> tag that vs added. I removed it and regenerated the file

docs/script-api.md
jeremypepper: fixed, it was the <summary> tags in comments visual studio added, I removed it.

src/game_api/script.cpp
Malacath-92: It's `00011110 00100000` which we could translate into what's pressed and not but off the top of my head I can't say  😅

Malacath-92: It's:
- Door pressed
- Whip released
- Bomb released
- Rope released
- Run released

Don't think that makes much sense

src/game_api/script.cpp
```diff
@@ -1037,6 +1037,27 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
         }
         return (uint16_t)0;
     };
+    /// Read input that has been previously stolen with steal_input
+    lua["read_stolen_input"] = [this](int uid)
+    {
+        Player* player = get_entity_ptr(uid)->as<Player>();
+        if (player == nullptr)
+            return (uint16_t)0;
+        ScriptInput* readinput = reinterpret_cast<ScriptInput*>(player->input_ptr);
```
Malacath-92, line 1046: So this cast is only valid if the input was actually stolen, you should check if that is the case before using this pointer.

Malacath-92 (changes requested):

src/game_api/script.cpp
jeremypepper: dumb question, is there a standardized way to check if the pointer is valid? do i use the IsBadReadPtr method?

jeremypepper: Not sure if this is the change you were looking for or if there is a way that I can read the pointer in a way to tell if it is valid or not. I ended up adding a boolean flag to tell if the input was stolen

src/game_api/script.cpp
jeremypepper: this is now unneeded by checking for the stolen input

src/game_api/script.cpp
```diff
@@ -1037,6 +1038,27 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
...
+        if (!IsBadReadPtr(readinput, 20))
+        {
+            if (!readinput->is_input_stolen)
```
Malacath-92, line 1050: So normally the game points to just a 32bit large struct, let's call it `GameInput`
```
struct GameInput {
    uint16_t next;
    uint16_t current;
};
```
Now when stealing the input OL allocates `ScriptInput` which shares a common initial sequence (aka the first two members have same alignment and size) And OL can access this pointer as a `ScriptInput` if and only if it has allocated it as a `ScriptInput`. What you are doing is to assume it always contains a `ScriptInput` when in reality it could contain a `GameInput`. So you might read past the end of the struct and read trash data. I know the intention is to only use this function when the input is stolen, but that's no good reason to assume it will only be called in that case.
So what you really need to do is check if `this.script_input` contains `uid`, if it does you treat the input as stolen, otherwise you treat it as vanilla.

Malacath-92 (changes requested):

src/game_api/script.cpp
jeremypepper: I think I fixed this? Can you take another look?

PR #77, opened 2021-05-21
Add 16 bytes of padding to audio according to the FMOD docs

Malacath-92:
This fixes a bug where fmod reads/writes out of the allocated buffer and sometimes that breaks audio, especially for short audio.


Dregu (approved):
If you say so! I've never even used this.

PR #79, opened 2021-05-24
Fix some docs bugs

Malacath-92:
/thread


Malacath-92:
@Dregu & @gmjosack I just wanted to unify the notation for functions because I didn't see a reason to have two ways to document a function. @jeremypepper is not a fan of this and would rather have the old style for both free and member functions so I just wanted to get your opinion before I act on Jeremy's feedback. See Discord for the short discussion.

gmjosack:
I personally prefer the previous style but prefer that we have one style of either over two styles

docs/script-api.md
```diff
@@ -1064,3 +1042,11 @@ Params: `PlayingSound vanilla_sound`
 - [`ICECAVES`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=COSUBTHEME.ICECAVES) 5
 - [`NEOBABYLON`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=COSUBTHEME.NEOBABYLON) 6
 - [`SUNKENCITY`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=COSUBTHEME.SUNKENCITY) 7
+## Aliases
+We use those to clarify what kind of values can be passed and returned from a function, even if the underlying type is really just an integer or a string. This should help to avoid bugs where one would for example just pass a random integer to a function expecting a callback id.
+### CALLBACK_ID == int;
```
jeremypepper, line 1047: should these be in all caps? all caps makes me think these are all constants

Malacath-92: I'm not sure about it, the enums we have are all caps and I wouldn't have done that so it's out of my comfort zone to determine what other types should or should not be all caps.

jeremypepper: IMO enums make sense as caps because their values are constants. Here we are defining types so I think these should be PascalCase

Dregu (approved):
I like the new docs format and the entity lists on github. Autocomplete seems to just work magically in vscode. New lua libs seem useful. Didn't check anything else you might have snuck in :)

gmjosack (approved):

Malacath-92:
I'll fix those conflicts later and merge.

PR #80, opened 2021-05-25
Adds script api: entity_get_items_by

jeremypepper:
entity_get_items_by accesses the inventory of a uid
with an item filter and a mask filter.
Added inventory.lua as an example.

### DIFF
--- a/examples/inventory.lua
+++ b/examples/inventory.lua
@@ -1,0 +1,42 @@
+meta.name = "Inventory Example"
+meta.version = "0.1"
+meta.description = "Examples for using player inventory."
+meta.author = "jeremyhay"
+
+local names = {}
+for i,v in pairs(ENT_TYPE) do
+  names[v] = i
+end
+
+-- display the items and the powerups in the player's inventory
+set_callback(function()
+    if players and players[1] then
+        x, y, l = get_position(players[1].uid)
+        -- you may want to mask only the items and logical (powerups)
+        -- here so you don't get any unwanted FX
+        -- note that ITEM_WHIP shows up while you are whipping
+        local ANY_ITEM = 0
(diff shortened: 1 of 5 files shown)


examples/inventory.lua
```diff
@@ -0,0 +1,42 @@
...
+
+-- display the items and the powerups in the player's inventory
+set_callback(function()
+    if players and players[1] then
+        x, y, l = get_position(players[1].uid)
+        -- you may want to mask only the items and logical (powerups)
+        -- here so you don't get any unwanted FX
+        -- note that ITEM_WHIP shows up while you are whipping
+        local ANY_ITEM = 0
+        local items = entity_get_items_by(players[1].uid, ANY_ITEM, MASK.ITEM & MASK.LOGICAL)
```
Malacath-92, line 19: What's `MASK.ITEM & MASK.LOGICAL`? If you want those two bits then you should do `MASK.ITEM | MASK.LOGICAL`

src/game_api/script.cpp
```diff
@@ -818,6 +818,8 @@ SpelunkyScript::ScriptImpl::ScriptImpl(std::string script, std::string file, Sou
     lua["entity_has_item_uid"] = entity_has_item_uid;
     /// Check if the entity `id` has some ENT_TYPE `type` in their inventory
     lua["entity_has_item_type"] = entity_has_item_type;
+    /// Check if the entity `id` has some ENT_TYPE `type` and `Mask` in their inventory. Set type and mask to 0 to return all inventory items.
```
Malacath-92, line 821: Maybe this should be more correctly `Gets all items of type and mask from entities inventory`

examples/inventory.lua
jeremypepper: you're right I don't know what I was thinking. Been a while since I used bitwise operators.

Malacath-92 (approved):

PR #81, opened 2021-05-25
Cleans up script api variable names

jeremypepper:
I found the script-api docs a bit confusing as uid, type, id, are all
kind of used interchangeably. This change attempts to normalize the
params.

functions expecting a uid will be named uid.
functions expecting a type will be named entity_type (named so users will associate this with a ENT_TYPE enum)
functions expecting a radius will be named radius (was just r)


jeremypepper:
I see the failures, looking

Malacath-92 (approved):

PR #84, opened 2021-05-26
drops, cosmic ocean theme and particle generation

zappatic:


docs/script-api.md
```diff
@@ -893,6 +905,14 @@ end, ON.LEVEL)
 - [`TITLE_TORCHFLAME_SMOKE`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PARTICLEEMITTER.TITLE_TORCHFLAME_SMOKE) 1
 - ...check particle_emitters.txt output by Overlunky...
 - [`MINIGAME_BROKENASTEROID_SMOKE`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PARTICLEEMITTER.MINIGAME_BROKENASTEROID_SMOKE) 219
+### DROPCHANCE
+- [`BONEBLOCK_SKELETONKEY`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=DROPCHANCE.BONEBLOCK_SKELETONKEY) 0
+- ...see_drops.hpp_for_a_list_of_possible_dropchances...
```
Malacath-92, line 910: Here it looks like you didn't need `_` for the docs to not break.

src/game_api/state.hpp
```diff
@@ -84,7 +90,13 @@ struct StateMemory
     Items* items;
     LevelGenSystem* level_gen;
     Layer* layers[2];
-    char pad12d0[0x1308 - 0x12d0];
+    size_t unknown23;
+    size_t unknown24;
+    size_t unknown25;
+    size_t unknown26;
+    size_t particle_emitters_info;
```
Malacath-92, line 97: Is this a pointer?

Malacath-92 (approved):

docs/script-api.md
zappatic: Do you mean inside the string? Like for_a_list etc? 
All the other comments like this, (including your sound_params :) ) do the same though. I'll check and see if it works without.

src/game_api/state.hpp
zappatic: It is, but I haven't investigated any deeper, and just needed this base pointer to give to the generate particles function. It's a pointer containing lots of pointers, several layers deeps, looks like they are maybe grouping particle emitters in sections or something. Will look into it more later.

docs/script-api.md
Malacath-92: This is another ultra-l33t-hax0r thing in the docs generator. It will replace double underscores in this context with spaces and we can't use spaces directly because some other part strips spaces but we also can't use single underscores because we use those for file names.

src/game_api/state.hpp
Malacath-92: That's fine, I was just curious  :)

PR #86, opened 2021-05-27
expose player inputs to lua

zappatic:


src/game_api/state.hpp
```diff
@@ -274,3 +311,52 @@ struct Illumination
...
+    uint8_t left;
+    uint8_t right;
+    uint8_t up;
+    uint8_t down;
+};
+
+struct PlayerSlot
+{
+    uint16_t buttons_gameplay;
+    uint16_t buttons;
```
Malacath-92, line 334: Does this have the same semantics as `Entity::buttons`? If so, what reason would one have to use this instead of `Entity::buttons`? If not, can we generalize the functions `is_button_pressed` and family to be usable with this?

jeremypepper:
Pulled and tested the script and the auto run detection works great, thanks for the change!

src/game_api/state.hpp
```diff
@@ -274,3 +311,52 @@ struct Illumination
...
+    uint8_t journal;
+    uint8_t left;
+    uint8_t right;
+    uint8_t up;
+    uint8_t down;
+};
+
+struct PlayerSlot
+{
+    uint16_t buttons_gameplay;
```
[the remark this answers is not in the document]
zappatic: I forgot to bring over the comments on these fields from the plugin. 'buttons' works all the time, 'buttons_gameplay' too, except in the game's pause menu.
If there are better names for these variables, let me know.

src/game_api/state.hpp
zappatic: Movable::buttons does not have up/down/left/right. The advantage of using movable is that you don't have to guess which player is associated with the buttons. If you ask the buttons of entity ana_spelunky for example, then that's what you get. 
If you access it through this new object in State, you have to loop all player slots to see which is correct. Or you'd have a function `is_button_pressed(button, playerslot)`

However, there is also Player::input_ptr, which is in fact exactly the same pointer as is in State! But, with the same restriction that it contains all player slots and you have to determine which slot you want.
I did not want to mess with Player::input_ptr, as it is already used in functions like steal_input etc, and the exact same object is now available through State anyways.

PR #88, opened 2021-06-02
Make set_contents documentation less confusing

jeremypepper:
The script docs seem to indicate that you need to pas a uid for
the item contents but we really want an entity id. A uid will cause
a crash. I'm also including the additional containers that are
supported with this function.


Dregu:
I think you made it confusing yourself in #81 :D

PR #92, opened 2021-06-09
Character definitions

Malacath-92:
Not gonna add this to the Lua API until we merge #90


gmjosack (approved):
This looks good to me though `set_heart_color` might be fun for lua to make dynamic hearts

Malacath-92:
Ah right, `set_heart_color` is actually the one that I wanted to expose. Not so much the name and gender stuff because that just seems wrong. But I'll take all y'all's lead on that  :)

Dregu:
I'm not sure lowercase f should be a reserved global in lua, seems kinda common... I liked F more.

Malacath-92:
Heh, that doesn't really matter to me. Just went with that python format strings would look like. I'll fix it later. If that's your only problem I'm even ready to merge.

Dregu:
I think that's all! I wasn't sure you could actually make Colors first, but turns out Color.new() is a thing!

Malacath-92:
All set now, added also `game_allocator` which might come in handy eventually, but for now it just avoids the crash in Debug when closing the game.

PR #96, opened 2021-06-11
Add spawn_liquid

Malacath-92:


src/game_api/rpc.cpp
```diff
@@ -27,6 +27,25 @@ uint32_t flipflag(uint32_t flags, int bit)
...
+    {
+        auto memory = Memory::get();
+        auto exe = memory.exe();
+        auto start = memory.after_bundle;
+        auto location = find_inst(exe, "\x41\xb9\x8b\x03\x00\x00\x0f\x28\xd7"s, start) - 0x1;
+        location = find_inst(exe, "\xe8"s, location);
+        location = decode_pc(exe, location, 1);
+        return memory.at_exe(location);
+    }
+    ();
```
Malacath-92, line 43: What the hell is this formatting clang?

Malacath-92:
@zappatic @Dregu @gmjosack I am very very unsure about the Lua interface I created for the spawn stuff. In particular the `set_pre_entity_spawn` lets you replace spawns which is very flexible. But that also means that users who use `spawn_entity` might get back an entity that they didn't ask for which can break the script easily. In addition, replacing some vanilla spawns just straight up breaks the game, for example replacing pets means the game interprets another entity as a pet, essentially writing random stuff into said entity. So I am very tempted to remove that interface and just make one `on_entity_spawn`, but I'd like some input on it.

Malacath-92:
Also, regarding the vtable hooks. If we want to bother with it we can make it much more efficient and less volatile if we require specifying the size of a vtable on hooking (from what I could see there was no way to programmatically determine that). Then we could just allocate the copy of the original vtable per entity and safely hook each function there, instead of having to hook the vtable for a while type and then redirecting based on the actual entity address. Even better for the dtor, which we right now have to hook the same call for each type, so all dtor calls are funneled through this one function. You get the gist  :)

Dregu:
> 
> 
> @zappatic @Dregu @gmjosack I am very very unsure about the Lua interface I created for the spawn stuff. In particular the `set_pre_entity_spawn` lets you replace spawns which is very flexible. But that also means that users who use `spawn_entity` might get back an entity that they didn't ask for which can break the script easily.

I don't mind keeping volatile functions in the api if they have their use cases and can be maintained, but I also see a lot of problems with this. Always catching all script spawn_entity()s too means creating a circular replacement just leads to a stack overflow. That should be optional at least.

Circumventing that I also tried making a little enemy rando with `set_pre_entity_spawn` and I'm just crashing all the time.
```lua
meta.name = 'Pre-Rando test'
meta.author = 'Dregu'

local script_spawn = false
local small = {ENT_TYPE.MONS_SNAKE, ENT_TYPE.MONS_SPIDER, ENT_TYPE.MONS_BAT, ENT_TYPE.MONS_CAVEMAN, ENT_TYPE.MONS_SKELETON, ENT_TYPE.MONS_SCORPION, ENT_TYPE.MONS_HORNEDLIZARD, ENT_TYPE.MONS_MANTRAP, ENT_TYPE.MONS_TIKIMAN, ENT_TYPE.MONS_MOSQUITO, ENT_TYPE.MONS_MAGMAMAN, ENT_TYPE.MONS_ROBOT, ENT_TYPE.MONS_CROCMAN, ENT_TYPE.MONS_COBRA, ENT_TYPE.MONS_JIANGSHI, ENT_TYPE.MONS_FEMALE_JIANGSHI, ENT_TYPE.MONS_OCTOPUS, ENT_TYPE.MONS_HERMITCRAB, ENT_TYPE.MONS_UFO, ENT_TYPE.MONS_YETI}

set_pre_entity_spawn(function(entity_type, x, y, layer, overlay_entity)
  if overlay_entity ~= nil then return end
  local new_uid = nil
  local new_type = small[math.random(#small)]
  if not script_spawn then
    script_spawn = true
    new_uid = spawn(new_type, x, y, layer, 0, 0)
    script_spawn = false
  end
  return new_uid
end, 0, small)
```

Malacath-92:
I'll try that script to see why it crashes.

As for making some spawns not be caught by the callbacks, I think that's a good idea. Perhaps add a `spawn_nonreplacable` or some such, which then just skips the `pre_entity_spawn` callback?

Dregu:
Just having `spawn_nonreplacable` would already make the rando script a lot simpler, but I might also want to apply a callback to original level gen entities only and ignore script entities, regardless of what the other script maker was thinking.

Then again having something like enemy randomizer work with an unrelated script that also adds enemies would be cool.

Malacath-92:
Hooking specific types of spawns will just take more research or hax, but it's probably a good idea in general. The tags I could think of off the top of my head are LEVEL_GEN, SCRIPTED and SYSTEMIC (though technically others might be relevant, like PLAYER, but need ever more research)
Personally I don't mind holding off with this until we have done the research, and in the meantime we merge only the liquid stuff.

Malacath-92:
@Dregu I added spawn_nonreplaceable and SPAWN_TYPE as an argument to the spawn callbacks

src/game_api/script/script_impl.cpp
```diff
@@ -153,12 +154,57 @@ ScriptImpl::ScriptImpl(std::string script, std::string file, SoundManager* sound
...
 
     /// Print a log message on screen.
-    lua["message"] = [this](std::string message) -> void
+    lua["print"] = [this](std::string message) -> void
     {
         messages.push_back({message, std::chrono::system_clock::now(), ImVec4(1.0f, 1.0f, 1.0f, 1.0f)});
         if (messages.size() > 20)
             messages.pop_front();
     };
+    /// Same as `message`
```
gmjosack, line 163: same as print?

zappatic (approved):

src/game_api/script/script_impl.cpp
```diff
@@ -642,6 +733,43 @@ ScriptImpl::ScriptImpl(std::string script, std::string file, SoundManager* sound
         return (uint16_t)0;
     };
 
+    /// Clears a callback that is specific to an entity.
+    lua["clear_entity_callback"] = [this](int uid, CallbackId cb_id)
+    {
+        clear_entity_hooks.push_back({uid, cb_id});
+    };
+    /// Returns unique id for the callback to be used in [clear_entity_callback](#clear_entity_callback) or `nil` if uid is not valid.
+    /// `uid` has to be the uid of a `Movable` or else stuff will break.
```
gmjosack, line 742: would it make sense to make the methods on Movable's then?

Malacath-92: Makes sense from that perspective, but the script has to keep track of which entities it hooked for cleanup on reload (so the movable doesn't call stale code), so moving the callback to `Movable` complicates things quite a bit

src/game_api/script/script_impl.cpp
Malacath-92: yes

gmjosack (approved):

PR #99, opened 2021-06-17
Reorganize + new stuff

zappatic:


Dregu:
The sound_pos and emitted_light in walltorchflame are common to torchflame and lampflame too. Curiously not other flames though... Why is there so many different flames!?

zappatic:
Mr. Auto had actually already looked at the different flames, of which I wasn't aware when I added this walltorchflame (I added it because of the request for changing the light color in the papercraft mod by Peabody) I'll make sure the class hierarchy is in order.

Mr-Auto:
well, seams like each flame is a little different

Malacath-92 (approved):

PR #103, opened 2021-06-26
Adds NaN checks to crashing imgui draw functions

jeremypepper:
New version of imgui crashes when a NaN is passed to one of the
arguments. The codepath that found this crash was to first measure
the text height when in fullscreen mode while alt-tabbed and then
feeding the output into draw_text.

Test Script:
```lua
set_callback(function()
    local x = 0
    local size = 24
    local text = "X"
    -- setting y directly won't cause a crash
    --local y = -0.5
    -- but using y as the result of a text_size causes a crash
    local text_w, text_height = draw_text_size(size, text)
    local y = text_height
    draw_rect(x, y, x+1, y, 2, 0, rgba(255,255,255,255))
end, ON.GUIFRAME)
```

Also updates imgui to latest in the docking branch

### DIFF
--- a/src/game_api/script/script_impl.cpp
+++ b/src/game_api/script/script_impl.cpp
@@ -153,13 +153,17 @@ ScriptImpl::ScriptImpl(std::string script, std::string file, SoundManager* sound
     /// Provides a read-only access to the save data, updated as soon as something changes (i.e. before it's written to savegame.sav.)
     lua["savegame"] = g_save;
 
+    /// Standard lua print function, prints directly to the console but not to the game
+    lua["lua_print"] = lua["print"];
     /// Print a log message on screen.
     lua["print"] = [this](std::string message) -> void
     {
         messages.push_back({message, std::chrono::system_clock::now(), ImVec4(1.0f, 1.0f, 1.0f, 1.0f)});
         if (messages.size() > 20)
             messages.pop_front();
+        lua["lua_print"](message);
     };
+
     /// Same as `print`
     lua["message"] = [this](std::string message) -> void
     { lua["print"](message); };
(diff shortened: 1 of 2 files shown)


Malacath-92:
Are you certain that just draw_rect is plagued by this issue?

jeremypepper:
That's a good point. I only know that draw_text and draw_rect_filled don't crash with a nan but I can do some more testing of other functions

jeremypepper:
Looks like it crashes in quite a few cases and warns about nan in others and silently does nothing in the text case.
```lua
local color = rgba(255,255,255,255)
local nan = 0/0

-- does not crash, prints nan error
--draw_line(nan,nan,nan,nan,nan,nan)

-- crash
--draw_rect(nan,nan,nan,nan, 1, 0, color)

-- crash
--draw_rect_filled(nan,nan,nan,nan, 0, color)

-- crash
 draw_circle(nan,nan,nan, 1, color)

-- does not crash, nan error
--draw_circle_filled(nan,nan,nan,nan, 1, color)

-- does not crash, does nothing silently
--draw_text(nan,nan,nan, "pls no crash", color)
```

jeremypepper:
I don't quite understand why sometimes we get the sol3 checker printing out an error about NaN and sometimes we don't

Malacath-92:
Considering that we receive nan only from the functions that adjust screen coordinates I think it's a better fix to test for nan there and return something more sensible. All other instances of nan are probably user error, right?

jeremypepper:
Even if the NaN is a user error I don't think we should crash. That's an easy mistake and it's going to be really hard for a user to track down, especially without an error or stack trace as to what is causing the crash.

Malacath-92:
Fair  :)
Then I advocate for adding them to all draw functions. We can investigate first how sol's checker doesn't always work. Perhaps there's a fix for that.

jeremypepper:
Ok, Added checks to the draw functions and verified with the script above.

Malacath-92:
I'm sorry to draw this out, but why don't you just add nan checks to all the draw functions? Just because some don't crash right now doesn't mean it's not UB anyways.

jeremypepper:
I guess because I was hoping it would be a temporary stopgap until we understand why sol's checker doesn't work. If I'm going to add it to all the functions I should probably print an error so the mod creator can fix there NaN. Could you point me how to do that?

PR #111, opened 2021-07-07
casts FLOOR_DOOR_ENTRANCE as floor

jeremypepper:
instead of as_door. This was throwing an error when I was looping
through floor tiles at the start of the level:

```lua
set_callback(function()
    local filtered_uids = {}
    local x,y,l = get_position(players[1].uid)
    local radius = 5
    local uids = get_entities_at(0, MASK.FLOOR, x, y, l, radius)
    for i, uid in ipairs(uids) do
        -- this commented out code works but the one below does not
        --if get_entity_raw(uid).type.id ~= ENT_TYPE.FLOOR_BORDERTILE then
        prinspect(i, uid, get_entity_type(uid))
        if get_entity_type(uid) == ENT_TYPE.FLOOR_DOOR_ENTRANCE then

        end
        if get_entity(uid).type.id ~= ENT_TYPE.FLOOR_BORDERTILE then
            table.insert(filtered_uids, uid)
        end
    end
    prinspect(filtered_uids)
end, ON.LEVEL)
```

I think maybe FLOOR_DOOR_ENTRANCE isn't actually a door. It doesn't go
anywhere after all and is maybe totally visual?

### DIFF
--- a/src/game_api/script/usertypes/entity_casting_lua.cpp
+++ b/src/game_api/script/usertypes/entity_casting_lua.cpp
@@ -31,7 +31,7 @@ void register_usertypes(sol::state& lua, ScriptImpl* script)
     lua["TYPE_MAP"][19] = lua["Entity"]["as_floor"]; // FLOOR_GROWABLE_VINE
     lua["TYPE_MAP"][20] = lua["Entity"]["as_floor"]; // FLOOR_CLIMBING_POLE
     lua["TYPE_MAP"][21] = lua["Entity"]["as_floor"]; // FLOOR_GROWABLE_CLIMBING_POLE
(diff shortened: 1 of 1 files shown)


Mr-Auto:
well, it is actually a door, but not the type that is in overlunky
https://github.com/spelunky-fyi/Spelunky2X64DbgPlugin/blob/master/resources/Spelunky2.json#L8490

zappatic:
Thanks for the script, that helped test it more easily. The solution to this problem wasn't to cast it as a Floor though, it is in fact a Door. The problem was that even though there was a C++ Door class, there was no Lua counterpart, and `get_entity` assumed there was. There used to be no reason for there to be a Lua class, as there were no parameters. This changed, as I added the ones found by Mr Auto

jeremypepper:
Cool, I'll close this PR then!

PR #113, opened 2021-07-07
set_max_velocity function

0novo:
Adds a function that sets the max velocity. The search methods I used are a little iffy I think

One of those max velocity variables has about 300 references (address 162277D3C in Ghidra, not sure if that's constant and/or the correct term :P). I'm not sure if this will cause any problems.
I have tested it a little and from what I've seen it works fine.

Edit: I think it's guaranteed to be problematic. MULSS, the instruction, takes m32 as its second operand. I can try putting in an offset to a static variable, but I don't know if that's possible and if it would work (nevermind, I just tried this and it didn't work). I'm too naive to be doing this stuff :/

### DIFF
--- a/src/game_api/rpc.cpp
+++ b/src/game_api/rpc.cpp
@@ -1331,3 +1331,26 @@ uint32_t waddler_entity_type_in_slot(uint8_t slot)
     }
     return 0;
 }
+
+void set_max_velocity(float max_positive_velocity, float max_negative_velocity)
+{
+    static size_t positive_offset = 0;
+    if (positive_offset == 0)
+    {
+        auto memory = Memory::get();
+        std::string pattern = "\xF3\x0F\x59\x15\x2A\x2A\x2A\x2A\xF3\x44\x0F\x10\x05\x2A\x2A\x2A\x2A"s;
+        positive_offset = memory.at_exe(find_inst(memory.exe(), pattern, memory.after_bundle) + 4);
+        // add 4 bytes since that's where the thing we want is
+    }
+    // instruction address + relative address thing + 4 bytes since it takes 4 bytes to say the relative address thing = magic
+    write_mem_prot(positive_offset + read_u32(positive_offset) + 4, to_le_bytes(max_positive_velocity), true);
+    // something like `x == NULL ? default_max_velocity : x` would be nice, not sure if that would work though
+    static size_t negative_offset = 0;
+    if (negative_offset == 0)
+    {
+        auto memory = Memory::get();
+        std::string pattern = "\xF3\x0F\x59\x0D\x2A\x2A\x2A\x2A\x41\x0F\x28\xD6\x48\xB8\x00\x00\x00\x00\x80\x01\x00\x00"s;
+        negative_offset = memory.at_exe(find_inst(memory.exe(), pattern, memory.after_bundle) + 4);
+    }
+    write_mem_prot(negative_offset + read_u32(negative_offset) + 4, to_le_bytes(max_negative_velocity), true);
+}
--- a/src/game_api/rpc.hpp
+++ b/src/game_api/rpc.hpp
@@ -84,3 +84,4 @@ void waddler_remove_entity(uint32_t entity_type, uint8_t amount_to_remove = 99);
 int16_t waddler_get_entity_meta(uint8_t slot);
 void waddler_set_entity_meta(uint8_t slot, int16_t meta);
 uint32_t waddler_entity_type_in_slot(uint8_t slot);
+void set_max_velocity(float max_positive_velocity, float max_negative_velocity);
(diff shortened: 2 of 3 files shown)


0novo:
I managed to allocate some memory within a 32-bit range of the instructions using `VirtualAllocEx`, meaning I was able to change the offset provided in the instructions to the allocated memory. The only problem that remains is that I'm getting an access violation. The .dmp file shows the exception code was `0xC0000005` and the message was `The thread tried to read from or write to a virtual address for which it does not have the appropriate access.` The indexing and instruction overriding works perfectly fine, so that is literally the only problem left to my knowledge (I really am worried about the possible side effects here, since it should do as it says and not have any side effects).
What the allocation code looks like currently:
```
    static float* velocity_pointer = nullptr;
    if (velocity_pointer == nullptr)
    {
        static HANDLE spel2_handle;
        if (spel2_handle == nullptr)
        {
            spel2_handle = [](void) -> HANDLE {
                HANDLE handle = nullptr;
                auto snapshot = CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, NULL);
                PROCESSENTRY32 process;
                process.dwSize = sizeof(PROCESSENTRY32);
                if (Process32First(snapshot, &process) == TRUE)
                {
                    while (Process32Next(snapshot, &process) == TRUE)
                    {
                        std::string exeName(process.szExeFile);
                        if (exeName == "Spel2.exe")
                        {
                            handle = OpenProcess(PROCESS_ALL_ACCESS, FALSE, process.th32ProcessID);
                            return handle;
                        }
                    }
                }
                if (!handle)
                {
                }
                return nullptr;
            }();
            // don't know how to do the part below yet
            //atexit([&spel2_handle, &velocity_pointer](void) -> void { VirtualFreeEx(spel2_handle, velocity_pointer, sizeof(float) * 2, MEM_RELEASE); };);
        }
        // todo: check if these the best settings, namely "PAGE_EXECUTRE_READWRITE". maybe find a way to not just add 0x10000000 as well
        velocity_pointer = (float*)VirtualAllocEx(spel2_handle, (LPVOID)(positive_offset + 0x10000000), sizeof(float) * 2, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
        if (velocity_pointer == nullptr) // allocation failed
        {
            DEBUG("error allocating memory for function `set_max_velocity`: {}", GetLastError());
        }
        else
        {
            velocity_pointer[0] = max_positive_velocity;
            velocity_pointer[1] = max_negative_velocity;
        }
    }
```
Obviously it could be improved quite a bit in all aspects, but I'm just focusing on getting it working.
I'm not currently aware of any methods to resolve this issue (the access violation), and I'm already pretty tired as well currently. Still, any suggestions and ideas are appreciated, and I can show some more code if that's needed to resolve this.

zappatic:
How many of these 300 references do you actually need to change? Is it just 1 location in the executable that needs to reference the two new floats? Or is it more like one per entity (so, quite a lot)?

I had a similar problem, where a value was hard-coded, but the const float that was referenced was used all over in the executable. Changing that float with many references made the game crash in other, totally unrelated parts.

That's why I did this dirty trick: https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/rpc.cpp#L852-L856

0novo:
> Is it just 1 location in the executable that needs to reference the two new floats?

Yeah, it's just one location (there might be another one in an "UndefinedFunction" which I saw in Ghidra, but it seems to work fine without changing that function).

> I had a similar problem, where a value was hard-coded, but the const float that was referenced was used all over in the executable. Changing that float with many references made the game crash in other, totally unrelated parts.

I mean, in this case it's more a problem of making sure nothing bad would happen from changing the float values. The game didn't crash or glitch out the last time I overwrote those floats, but that was just at a glance.

> That's why I did this dirty trick: https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/rpc.cpp#L852-L856

So you're putting the float in the "0xCC" stuff between the functions, and then overwriting the instructions to reference that float? Is that what's happening there?

zappatic:
That is correct

Dregu:
To be clear, this changes the terminal velocity of 0.4 of the players and monsters or what? "The max velocity" doesn't really mean anything.

0novo:
> To be clear, this changes the terminal velocity of 0.4 of the players and monsters or what? "The max velocity" doesn't really mean anything.

You're right. "Global max velocity" would be a more specific name. It's a hard cap on anything that can have velocity

Dregu:
> You're right. "Global max velocity" would be a more specific name. It's a hard cap on anything that can have velocity

It's not global though, projectile velocity (bullet, laser...) is not capped I think.

0.4 applies to most things like monsters and rocks.

0novo:
I got the 0xCC overriding to work, but it turns out that the downwards velocity cap isn't actually decided in that one function.
There seems to be multiple functions (some "undefined"?) that control falling logic and use the downwards velocity cap (0.4). I don't know whether they are virtual functions or not, but I don't think I will spend any more time on getting that to work (I'm not even that knowledgeable in reverse-engineering). After all, I only started this because of a pretty dumb mod idea (ABHing in Spelunky 2).
The other axes work fine, I could just remove the negative y-velocity argument (or leave it.)

As for other more specific/accurate names/descriptions for this function... yeah, I don't have anything right now. It's probably better to just leave this in PR limbo or to close it if there's not any suitable name

PR #117, opened 2021-07-11
Gui draw context

Malacath-92:
Make sure people never use draw functions outside of guiframe again (wishful thinking)


src/game_api/entity.cpp
```diff
@@ -402,22 +402,60 @@ void Entity::destroy()
...
+    {
+        float vx{0.0f};
+        float vy{0.0f};
+        if (ent->type->search_flags & 0b1111) // PLAYER | MOUNT | MONSTER | ITEM
+        {
+            Movable* mov = ent->as<Movable>();
+            vx = mov->velocityx;
+            vy = mov->velocityy;
+        }
+        if (ent->overlay)
```
Mr-Auto, line 433: shouldn't this be a loop to determinate top entity?
like when you on a mount and elevator, it will return 0. And it can goo even further if you check speed of jetpack

also fx and activefloor are also movable

Malacath-92: See line 435

Malacath-92: Will fix fx and activefloor, thanks for that  👍

PR #118, opened 2021-07-13
fixing bugs in the API, changed names to unique in it's hierarchy

Mr-Auto:


src/game_api/entities_items.cpp
```diff
@@ -0,0 +1,35 @@
...
+            flags = flags ^ 0b1001; // unset 1 & 4
+        if ((more_flags & 0b1000) != 0b1000)
+            more_flags = more_flags | 0b1000; // set 4
+        if (y == 0.0)
+            y = 0.5;
+        offsety = 0.115;
+        hitboxy = 0.285;
+        duckmask = 2143944961; // hehe funny number
+    }
+}
```
Malacath-92, line 35: Can you add a new line at the end of file please.

src/game_api/entities_items.hpp
```diff
@@ -36,7 +36,7 @@ class Bomb : public Movable
     size_t unknown1;
     float scale_hor; // 1.25 = default regular bomb ; 1.875 = default giant bomb ; > 1.25 generates ENT_TYPE_FX_POWEREDEXPLOSION
     float scale_ver;
-    bool unknown2;
+    bool is_big_bomb; // is bomb from powerpack
```
Malacath-92, line 39: If you add new comments, try placing them using the usual `///` in front of the member so the docs pick it up.

src/game_api/script/usertypes/entities_items_lua.cpp
```diff
@@ -518,13 +527,6 @@ void register_usertypes(sol::state& lua, ScriptImpl* script)
         sol::base_classes,
         sol::bases<Entity, Movable>());
 
-    lua.new_usertype<Container>(
```
Malacath-92, line 521: Why was this removed?

Malacath-92: Nevermind, it was double. Had to extend the context to see that.

Malacath-92 (changes requested):

src/game_api/entities_items.hpp
Mr-Auto: oh, that's a thing? maybe i should edit more of the comments that way?

Malacath-92: I think it should work. And yes, they should be in front.
If you want to you are free to change more of them, but there's no need to do that right now, so you can focus on something you enjoy more  :)

Malacath-92 (approved):

PR #121, opened 2021-07-19
Repl console

### LINKED ISSUE #119: Add a REPL console
Something that can be interfaced with from the dll interface so it can be disabled when in user mode (for PL)
I guess it would need to be its own Lua VM but not sure how to achieve it exactly.

Malacath-92:
Adds a console to close #119 
Current features:
- per-session history
- register_console_command from scripts
- deep tab-completion, aka it tries to correctly complete code like `player.overlay.he` by inspecting the metatable of `player.overlay`
- behaves like a script that is always loaded and enabled, so callbacks and all that stuff should work
- x-session history
- history shared with Playlunky
- configurable history length 

Known issues:
- the key to open the console is not in the same place on all keyboards
- tab-completion of array members doesn't work, e.g. `players[1].h` will attempt to tab-complete `.h`

Potential additional features:
- access previous results through shortcuts (for example $1 gives you the last result)


Dregu:
This is looking great! Then the complaints:

- The console is on top of other OL windows, making OL unclickable. The whole console apart from the input field should be click-through too to use mouse tools while the console is open, since there's nothing to interact with anyway. (This is kinda funky cause the clickhandler window should be over the console, but not the input field.)
- Can't press back down to empty line after pressing up to history.
- Could use a way to reset the script environment, to clear all callbacks and dumb stuff you've done

Non-console related stuff from lua refactoring:
- Rando doesn't work anymore, something with set_interval in required files?
```
[debug] stack index 3, expected string, received no value: bad get from protected_function_result (is not an error) 
stack traceback:
        [C]: in function 'base.set_interval'
        ...s/Dregu/Documents/Code/overlunky/examples/rando\door.lua:202: in function <...s/Dregu/Documents/Code/overlunky/examples/rando\door.lua:183>
[debug] stack index 3, expected string, received no value: bad get from protected_function_result (is not an error)
stack traceback:
        [C]: in function 'base.set_timeout'
        .../Dregu/Documents/Code/overlunky/examples/rando\enemy.lua:117: in function <.../Dregu/Documents/Code/overlunky/examples/rando\enemy.lua:115>
```
- A bunch of other stuff with `require` is also broken.

PR #126, opened 2021-07-25
New UI stuff

Mr-Auto:


src/game_api/rpc.cpp
```diff
@@ -60,17 +57,17 @@ float get_zoom_level()
...
-    auto state = State::get();
-    auto player = state.items()->player(0);
-    if (player == nullptr)
-        return;
-    for (auto& item : state.layer(player->layer)->items())
-    {
-        DEBUG("Item {} {:x}, {}", item->uid, item->type->search_flags, item->position_self());
-    }
-}
+//void list_items()
```
Malacath-92, line 60: You wanna remove this code if it's not used?

src/game_api/rpc.cpp
```diff
@@ -920,7 +917,7 @@ void set_kapala_hud_icon(int8_t icon_index)
 
     if (instruction_offset == 0)
     {
-        auto state = State::get();
+        const auto& state = State::get();
```
Malacath-92, line 920: Did you run some sort of clang-tidy on this?

src/game_api/rpc.hpp
```diff
@@ -8,7 +8,7 @@ void teleport(float x, float y, bool s, float vx, float vy, bool snap);
 void godmode(bool g);
 void darkmode(bool g);
 void zoom(float level);
-void list_items();
+//void list_items();
```
Malacath-92, line 11: See above

src/game_api/rpc.cpp
Mr-Auto: oh, i forgot, i commented that to see if it will fail to build

i guess is not needed? seams like pure debug leftover stuff

src/game_api/rpc.cpp
Mr-Auto: it was throwing warning (or check? they are all called warnings in MSVC i guess) C26820

Mr-Auto: also the state variable is not used here, does the get() function does something else? maybe i should remove it completely?

Mr-Auto: i will fix it more in my warning fix PR, this does nothing anyway, can pass for now

src/injected/ui.cpp
```diff
@@ -3203,6 +3277,7 @@ void render_entity_props()
     if (ImGui::CollapsingHeader("Stats"))
     {
         ImGui::DragScalar("Health##EntityHealth", ImGuiDataType_U8, (char*)&g_entity->health, 0.5f, &u8_one, &u8_max);
+        ImGui::DragScalar("Price##Price", ImGuiDataType_U8, (char*)&g_entity->price, 0.5f, &u32_zero, &u32_max);
```
Dregu, line 3280:
```suggestion
        ImGui::DragScalar("Price##Price", ImGuiDataType_S32, &g_entity->price, 0.5f, &s32_min, &s32_max);
```

PR #127, opened 2021-07-28
Adding and removing powerups

zappatic:
+ reorganize player to entities_chars*
+ show powerups in OL Special Attributes


Mr-Auto:
can you also expose the powerups so we can check if player has powerup?

also does it only work for player? i know that vampires and vlad have kapala for example

PR #128, opened 2021-07-29
Add TeleporterBackpack (aka Telepack)

mauvealerts:
I assume this is the same size as the Teleporter field


Dregu:
Oh yeah you need to add a few things to https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/script/usertypes/entities_items_lua.cpp#L107 to expose the usertype to lua.

Mr-Auto:
and https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/script/usertypes/entity_casting_lua.cpp so get_entity will return proper type

Malacath-92:
That entity casting stuff is auto generated. We should probably refrain from manually changing it and instead add the type to the x64dbg plugin.

mauvealerts:
I think I've done everything. Let me know if there's more

PR #129, opened 2021-08-03
Pushing overlunky to new standards

Mr-Auto:
Current problems:
- ~~global table players don't work, not sure why~~
- ~~\overlunky\src\toml11\toml\parser.hpp is being compiled with W4 for some reason~~


Malacath-92:
Regarding `toml11\toml\parser.hpp` containing warnings, it's a header so it is compiled with whatever TranslationUnit includes it. If you have warnings in there wrap the include in `#pragma warning(push, 0)` and `#pragma warning(pop)`

Mr-Auto:
@Malacath-92 it's included as private instead of public, i was hoping changing that will solve the issue but not :/
i guess i go with the pragma for now, also we have this library only for like on function in ui.cpp, do we really need it?

Malacath-92:
If we wanna read toml files we need a lib that reads toml files  :)
I don't think it's worth to parse them ourselves.

Dregu:
There are other toml libraries if you want to try if they have less warnings. 😛

The current one was chosen because we were using c++11 then, [this one](https://github.com/marzer/tomlplusplus) is for 20.

Dregu:
> global table players don't work, not sure why

Also I have no idea what you mean.

docs/script-api.md
```diff
@@ -225,7 +225,7 @@ Add a button that the user can click in the UI. Sets the timestamp of last click
 Spawn a "block" of liquids, always spawns in the front layer and will have fun effects if `entity_type` is not a liquid.
 Don't overuse this, you are still restricted by the liquid pool sizes and thus might crash the game.
 ### [`spawn_entity`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=spawn_entity)
-`int spawn_entity(int entity_type, float x, float y, int layer, float vx, float vy)`<br/>
+`int spawn_entity(int entity_type, float x, float y, int enum_layer, float vx, float vy)`<br/>
```
Malacath-92, line 228: Instead of renaming them all maybe add a typedef for it and specify the parameter as that. See for example what we do with `VANILLA_SOUND`

src/CMakeLists.txt
```diff
@@ -119,23 +138,38 @@ if(BUILD_OVERLUNKY)
...
-                        add_custom_target(update_auto_docs_overlunky
-                                COMMAND cd docs && ${Python3_EXECUTABLE} generate.py
-                                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-                                COMMENT "Update only auto-docs."
-                        )
-                else()
-                        message("You need python 3 for auto-docs.")
+        if(MSVC)
+                target_compile_options(injected PRIVATE 
+                /W4 /WX
```
Malacath-92, line 143: Formatting is quite off here

src/game_api/CMakeLists.txt
```diff
@@ -20,5 +20,11 @@ target_precompile_headers(spel2_api PRIVATE
         <d3d11.h>)
 
 if (MSVC)
-        target_compile_options(spel2_api PRIVATE /bigobj)
+    target_compile_options(spel2_api PRIVATE
```
Malacath-92, line 23: Indentation doesn't fit with the rest of the file

src/game_api/file_api.hpp
```diff
@@ -14,7 +14,7 @@ struct FileInfo
     void* Data{nullptr};
     int _member_1{0};
     int DataSize{0};
-    int AllocationSize{0};
+    long long int AllocationSize{0};
```
Malacath-92, line 17: Are you sure this is right? It changes the size of the struct.

src/game_api/level_api.cpp
```diff
@@ -920,13 +920,13 @@ std::pair<int, int> LevelGenSystem::get_room_index(float x, float y)
         static_cast<int>(std::ceil(x - 3.5f)) / 10,
         static_cast<int>(std::ceil(121.5f - y)) / 8};
 }
-std::pair<float, float> LevelGenSystem::get_room_pos(int x, int y)
+std::pair<float, float> LevelGenSystem::get_room_pos(unsigned int x, unsigned int y)
```
Malacath-92, line 923: Please use uint32_t instead of unsigned int (we should also use int32_t imho but I think we're past that)

src/game_api/rpc.cpp
```diff
@@ -232,7 +218,7 @@ uint32_t get_entity_flags(uint32_t uid)
 
 void set_entity_flags(uint32_t uid, uint32_t flags)
 {
-    if (uid == 0)
+    if (uid < 1)
```
Malacath-92, line 221: Why make this change?

src/game_api/script/lua_vm.cpp
```diff
@@ -990,8 +993,8 @@ std::shared_ptr<sol::state> acquire_lua_vm(class SoundManager* sound_manager)
     {
         assert(sound_manager != nullptr && "SoundManager needs to be passed to first call to get_lua_vm...");
 
-        std::shared_ptr<sol::state> global_vm = std::make_shared<sol::state>();
-        sol::state& lua_vm = *global_vm;
+        std::shared_ptr<sol::state> global_vms = std::make_shared<sol::state>();
```
Malacath-92, line 996: What is `vms` supposed to me?

src/game_api/script/usertypes/entity_casting_lua.cpp
```diff
@@ -1,5 +1,4 @@
 #include "entity_casting_lua.hpp"
-
```
Malacath-92, line 2: There's no need to remove empty lines, those kinds of things are fixed by clang-format and we don't enforce it. So whoever wrote this file might've had a good reason to put the empty line. Afaik empty lines after the corresponding header include is very standard.

src/game_api/script/usertypes/gui_lua.cpp
```diff
@@ -338,7 +338,7 @@ void register_usertypes(sol::state& lua)
     /// Converts a color to int to be used in drawing functions. Use values from `0..255`.
     lua["rgba"] = [](int r, int g, int b, int a) -> uColor
     {
-        LuaBackend* backend = LuaBackend::get_calling_backend();
+        //LuaBackend* backend = LuaBackend::get_calling_backend(); //???
```
Malacath-92, line 341: Delete, probably a leftover from automated refactoring

src/game_api/sound_manager.cpp
```diff
@@ -562,7 +562,7 @@ CustomSound SoundManager::get_event(std::string_view event_name)
     }
     return CustomSound{nullptr, nullptr};
 }
-PlayingSound SoundManager::play_event(FMODStudio::EventDescription* fmod_event, bool paused, bool as_music)
```
Malacath-92, line 565: Don't remove this, make a TODO instead that this needs to be fixed

src/game_api/vtable_hook.hpp
```diff
@@ -6,7 +6,7 @@
 #include <unordered_map>
 
 void* register_hook_function(void*** vtable, size_t index, void* hook_function);
-void unregister_hook_function(void*** vtable, size_t index, void* hook_function);
+//void unregister_hook_function(void*** vtable, size_t index, void* hook_function);
```
Malacath-92, line 9: Please keep this and just tag it as [[maybe_unused]]

src/injected/main.cpp
```diff
@@ -39,22 +39,26 @@ BOOL WINAPI ctrl_handler(DWORD ctrl_type)
 
 void attach_stdout(DWORD pid)
 {
-    //if (std::getenv("OL_DEBUG"))
+    //size_t requiredSize;
+    //getenv_s(&requiredSize, NULL, 0, "OL_DEBUG");
+    //if (requiredSize != 0)
```
Malacath-92, line 44: This should not be commented, right? I think I forgot to uncomment it before I merged

Malacath-92: What's with the naming though?

Mr-Auto:
> There are other toml libraries if you want to try if they have less warnings. stuck_out_tongue
> 
> The current one was chosen because we were using c++11 then, [this one](https://github.com/marzer/tomlplusplus) is for 20.

well, imgui has hundreds of warnings, but they are not a problem, so i dont understand why the compiler takes these from tomla into account

src/game_api/rpc.cpp
Mr-Auto: i didn't saw it's already unsigned xd 
i was actually thinking do we even need that check? if you give it wrong uid it will just not find it right?

Malacath-92:
@Mr-Auto I explained why those warnings in the toml headers are triggered. Warnings in header files will be triggered by the file that includes them.

src/injected/main.cpp
Mr-Auto: i don't really understand the _s variant of this function, this is copied from some example

src/game_api/script/lua_vm.cpp
Mr-Auto: just changed the name cuz of static above

src/injected/main.cpp
Malacath-92: You'll want something along the lines of 
```C++
size_t env_var_size;
getenv_s(&env_var_size, NULL, 0, "OL_DEBUG");
if (env_var_size > 0) // Means that the env var is set
```
Since you don't care about the value of the env var you just care whether it exists or not.

PR #130, opened 2021-08-03
Even more level gen stuff

Malacath-92:
Adds:
- room_template argument to tile code hooks
- spawn_tree for spawning a tree instead of just the stump
- define_room_template for adding new room templates
- override/add_level_files for loading new level files
- extra spawn API for semi-guaranteed spawns à la ghost pot or pet
- the beginnings of a prng API
- add shop_type to LevelGen

Fixes:
- decoration for SURFACE_HIDDEN floor
- add TUNNEL_CURRENT and TUNNEL_NEXT to the fix_decorations code
- spawn type flags were set totally wrong
- avoid the manual cast that is necessary for post_entity_spawn and pre/post_statemachine


src/game_api/prng.cpp
```diff
@@ -0,0 +1,33 @@
...
+	const auto& state = State::get();
+	static PRNG* prng = (PRNG*)((size_t)state.ptr_local() - 0xb0);
+	return *prng;
+}
+
+std::size_t PRNG::random_index(std::size_t size, PRNG_CLASS type)
+{
+	prng_pair& pair = pairs[type];
+
+	const std::uint64_t lower = pair.first;
```
Malacath-92, line 22: @zappatic This code seems to generate almost identical assembly as the game contains, only with different instruction ordering. But I thought that it shouldn't actually matter, right? As long as we do something deterministic here and touch the same pairs as the corresponding code in the game would it should all be fine.

zappatic: Yes, the order shouldn't really matter, because the two things happening (advancing the pair and calculating the random number) are separate from each other, as the random number generation happens on the already existing pair value, not the advance.
I assume the Ghidra C++ equivalent of the advance is correct. Unless there is some very specific math behind them, like using specific primes or something, I don't believe it really matters which 64 bit value you end up after the advance anyway, as long as it is big enough. Looks good and deterministic indeed.

docs/script-api.md
```diff
@@ -275,6 +277,9 @@ Short for [spawn_layer_door](#spawn_layer_door).
 ### [`spawn_apep`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=spawn_apep)
 `int spawn_apep(float x, float y, LAYER layer, bool right)`<br/>
 Spawns apep with the choice if it going left or right, if you want the game to choose use regular spawn functions with `ENT_TYPE.MONS_APEP_HEAD`
+### [`spawn_tree`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=spawn_tree)
```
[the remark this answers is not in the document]
Malacath-92: Because someone asked for this specifically and I quickly found a function in the game that does it. Dregu and me discussed on Discord to make a general function that does this kinda stuff based on the entity type, but right now we don't have much more.

PR #131, opened 2021-08-07
More api additions

zappatic:


src/game_api/render_api.cpp
```diff
@@ -137,3 +139,49 @@ const char** RenderAPI::load_texture(std::string file_name)
...
+void init_render_hud_hook()
+{
+    static size_t func_render_hud = 0;
+
+    auto& memory = Memory::get();
+    auto exe = memory.exe();
+
+    if (func_render_hud == 0)
+    {
+        std::string pattern = "\x48\x8D\x0D\x56\x05\x16\x00"s;
```
Malacath-92, line 166: I'm confused, what is the RVO of this function? Because if I use this pattern I find a function that def has more parameters than just one pointer.

zappatic: The function is at `221C1E90`. There's two extra parameters indeed, two floats for the y-level of the hud and its transparency. It doesn't really matter though, because the trampoline is the first function that gets called in the custom render_hud function, so `xmm1` and `xmm2` aren't touched and are passed on to the native function. I did add the params though, to be correct.
If you're looking at what Ghidra thinks the signature is, then as far as I can tell that's just wrong. If you look at where this function is called from, one level up in the call stack, you see `rcx`, `xmm1` and `xmm2` being prepared, that's all.

Malacath-92: I get your point, but imho it's better if we match the signature correctly. So in case we'll want to do anything else before calling the trampoline we don't suddenly break things without having an obvious reason.

And I know Ghidra is utterly useless with functions that take arguments in xmm registers, I got used to patching the signatures all the time  😬

src/game_api/script/lua_vm.cpp
```diff
@@ -661,6 +661,12 @@ end
     lua["waddler_set_entity_meta"] = waddler_set_entity_meta;
     /// Gets the entity type of the item in the provided slot
     lua["waddler_entity_type_in_slot"] = waddler_entity_type_in_slot;
+    /// Draw text on the screen using the built-in renderer. Use in combination with ON.VANILLA_RENDER
+    lua["vanilla_draw_text"] = vanilla_draw_text;
+    /// Measure the provided text using the built-in renderer. Returns the result as if the string was rendered with vanilla_draw_text.
+    lua["vanilla_measure_text"] = vanilla_measure_text;
```
Dregu, line 667: The gui version is called _draw_text_size_, might want to stay consistent here too with vanilla_draw_text_size...

Malacath-92:
Also curious if there are timing restrictions on calling the draw functions? If it is similar to the imgui functions I think it's worth to do as the imgui functions do and allow to call them only from a context that is passed to `ON.VANILLA_RENDER`
Also `ON.VANILLA_RENDER` should probably called on `ON.VANILLA_RENDER_GUI` or am I misinterpreting stuff? Is it only HUD or all the GUI?

src/game_api/render_api.cpp
```diff
@@ -137,3 +139,198 @@ const char** RenderAPI::load_texture(std::string file_name)
...
+
+    LuaBackend::for_each_backend(
+        [&](LuaBackend& backend)
+        {
+            backend.vanilla_render();
+            return true;
+        });
+}
+
+void init_render_hud_hook()
```
Malacath-92, line 157: I would rename this to `init_render_api_hooks` since we'll probably add other hooks here later.

src/game_api/render_api.cpp
```diff
@@ -137,3 +139,198 @@ const char** RenderAPI::load_texture(std::string file_name)
...
+{
+    static size_t func_render_hud = 0;
```
Malacath-92, line 159: Also all this stuff does not need to be static, the function is called only once

src/game_api/render_api.cpp
Malacath-92: In fact I just added more hooks and I'll probably push it later today. Sorry  😬

src/game_api/script/usertypes/vanilla_render_lua.hpp
```diff
@@ -0,0 +1,30 @@
...
+    void draw_text(const std::string& text, float x, float y, float scale_x, float scale_y, Color color, uint32_t alignment, uint32_t fontstyle);
+
+    /// Measure the provided text using the built-in renderer
+    std::pair<float, float> draw_text_size(const std::string& text, float scale_x, float scale_y, uint32_t fontstyle);
+
+    /// Draw a texture in screen coordinates `x`, `y` using the built-in renderer. Use in combination with ON.RENDER_PRE_HUD or ON.RENDER_POST_HUD
+    void draw_texture(uint32_t texture_id, uint8_t row, uint8_t column, float render_at_x, float render_at_y, float render_width, float render_height, Color color);
+
+  private:
+    class LuaBackend* backend;
```
Malacath-92, line 24: This is not needed, right?

zappatic: Right, no need for that indeed

src/game_api/render_api.cpp
```diff
@@ -142,6 +143,197 @@ const char** RenderAPI::load_texture(std::string file_name)
...
+            backend.post_render_hud();
+            return true;
+        });
+}
+
+using VanillaRenderPauseMenuFun = void(float*);
+VanillaRenderPauseMenuFun* g_render_pause_menu_trampoline{nullptr};
+void render_pause_menu(float* drawing_info)
+{
+    LuaBackend::for_each_backend(
```
Malacath-92, line 172: Maybe move this to events.hpp/.cpp because of compile times? lua_backend.hpp is quite heavy.

src/game_api/script/lua_backend.cpp
```diff
@@ -713,6 +716,156 @@ void LuaBackend::post_entity_spawn(Entity* entity, int spawn_type_flags)
...
+        }
+    }
+}
+void LuaBackend::pre_render_pause_menu()
+{
+    if (!get_enabled())
+        return;
+
+    /// Runs before the pause menu is drawn on screen. You can draw text and textures on screen by using the provided render_ctx parameter
+    sol::optional<sol::function> on_render_pre_pause_menu = lua["on_render_pre_pause_menu"];
```
Malacath-92, line 775: Are we sure we want to add more of those kinds of callbacks? We said in the past that they are a bit too magic so we stick to `set_callback`. It would also make the code a tiny bit less complex.

src/game_api/script/lua_backend.cpp
```diff
@@ -713,6 +716,156 @@ void LuaBackend::post_entity_spawn(Entity* entity, int spawn_type_flags)
...
+    {
+        auto now = get_frame_count();
+        if (callback.screen == ON::RENDER_POST_HUD)
+        {
+            handle_function(callback.func, render_ctx);
+            callback.lastRan = now;
+        }
+    }
+}
+void LuaBackend::pre_render_pause_menu()
```
Malacath-92, line 769: And also all those draw functions look extremely similar, any chance we can unify them?

src/game_api/script/lua_backend.cpp
```diff
@@ -713,6 +716,156 @@ void LuaBackend::post_entity_spawn(Entity* entity, int spawn_type_flags)
...
+    VanillaRenderContext render_ctx;
+
+    if (on_render_post_pause_menu)
+    {
+        on_render_post_pause_menu.value()(render_ctx);
+    }
+
+    for (auto& [id, callback] : callbacks)
+    {
+        auto now = get_frame_count();
```
Malacath-92, line 811: Shouldn't this be placed outside of the loop?

docs/script-api.md
```diff
@@ -2666,6 +2688,16 @@ Pop unique identifier from the stack. Put after the input.
...
+### `VanillaRenderContext`
+- [`nil draw_text(const string& text, float x, float y, float scale_x, float scale_y, Color color, int alignment, int fontstyle)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=draw_text) &VanillaRenderContext::draw_text
+\
+Draw text using the built-in renderer. Use in combination with ON.RENDER_PRE_HUD or ON.RENDER_POST_HUD. See vanilla_rendering.lua in the example scripts.
+- [`tuple<float, float> draw_text_size(const string& text, float scale_x, float scale_y, int fontstyle)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=draw_text_size) &VanillaRenderContext::draw_text_size
+\
+Measure the provided text using the built-in renderer
+- [`nil draw_texture(int texture_id, int row, int column, float render_at_x, float render_at_y, float render_width, float render_height, Color color)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=draw_texture) &VanillaRenderContext::draw_texture
+\
+Draw a texture in screen coordinates `x`, `y` using the built-in renderer. Use in combination with ON.RENDER_PRE_HUD or ON.RENDER_POST_HUD
```
Malacath-92, line 2700: This comment should probably just say `ON.RENDER_*` at this point since you added a million new events  :D

Malacath-92:
Maybe we can add some common colors to the API?`

src/game_api/render_api.cpp
zappatic: I did this, but the stuff I saw in events.cpp was more similar to the functions in render_api.cpp, not lua_backend.cpp. I did reduce the amount of functions in lua_backend.cpp from six to two though

zappatic:
Yes, I can save the bounds inside the VanillaRenderContext upon its creation, and check for overlap when draw_texture gets called. I could do the same for text, but that would mean I'd always have to do an extra draw_text_size.

Malacath-92:
> Yes, I can save the bounds inside the VanillaRenderContext upon its creation, and check for overlap when draw_texture gets called. I could do the same for text, but that would mean I'd always have to do an extra draw_text_size.

Yeah, I see the issue with text. Maybe we can evaluate down the line how big of a perf impact one or the other solution would have.

Malacath-92 (approved):

Issue #133, opened 2021-08-09
PRNG functions


Dregu:
I don't know how much sense starting random_index from 0 makes in lua...
random_int crashes when min==max and does crazy stuff when min>max

Anyway adding prng:random() that works as a drop in replacement for math.random() would be neat. I don't know if you already had this in mind.

> math.random() with no arguments generates a real number between 0 and 1.
> math.random(upper) generates integer numbers between 1 and upper (both inclusive).
> math.random(lower, upper) generates integer numbers between lower and upper (both inclusive).

Malacath-92:
Oh right, how did I not think of that 🤦‍♀️
I'll look into those

Malacath-92:
@Dregu Do you know what the other pairs are used for or did you just pick them randomly in rando2?

Dregu:
> did you just pick them randomly in rando2?

As is tradition. Well actually all I know is that some of them only change on level gen, others when you blow up stuff and some on every frame. I tried to pick appropriate ones for the occasion but probably forgot what I was supposed to do halfway.

Malacath-92:
I looked at some level gen code and all I saw used the first pair, hence I labeled it as such. I guess I'll have to look at more code to figure out at least some of the others.

PR #134, opened 2021-08-10
Improve the prng interface

Malacath-92:


Malacath-92:
@Dregu re #133 
This is not a drop-in replacement yet because you still have to pass in which pair to use and I have zero idea what would be a good default.

PR #136, opened 2021-08-13
Doc fix

Mr-Auto:
fix few things 
moved get_entity_ai_state to the Deprecated Functions course of the inconsistent name
i was thinking about adding get_entity_move_state for quick replacement, what you think?


Mr-Auto:
i take suggestions for better names then (set/get)_correct_ushabti_af :smile:

src/game_api/script/lua_vm.cpp
```diff
@@ -885,6 +894,8 @@ end
         return sol::nullopt;
     };
 
+    lua.create_named_table("BUTTONS", "NONE", 0, "JUMP", 1, "WHIP", 2, "BOMB", 4, "ROPE", 8, "RUN", 16, "DOOR", 32, "MENU", 64, "JOURNAL", 128, "LEFT", 256, "RIGHT", 512, "UP", 1024, "DOWN", 2048);
```
Malacath-92, line 897: This should probably be called `INPUT` to mirror the usage with the functions named `*_input` and avoid confusion with the enum `BUTTON`

src/game_api/script/usertypes/state_lua.cpp
```diff
@@ -141,7 +141,11 @@ void register_usertypes(sol::state& lua)
         "items",
         &StateMemory::items,
         "camera_layer",
-        &StateMemory::camera_layer);
+        &StateMemory::camera_layer,
+        "get_correct_ushabti_af",
+        &StateMemory::get_correct_ushabti_af,
+        "set_correct_ushabti_af",
+        &StateMemory::set_correct_ushabti_af);
```
Malacath-92, line 148: How about just writing it out in full so no one has to guess what `af` means?

Mr-Auto: I'm not a fan of long names for functions :/

Malacath-92: If you don't want long function names then remove the `_af` suffix and properly document it

src/game_api/script/lua_vm.cpp
Mr-Auto: sure, but INPUT is already define by some windows lib, so can't make it a type for the `*_input` functions :/

Dregu: You can certainly make it a lua enum though

src/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(MSVC)
                 target_compile_options(overlunky_warnings INTERFACE -Wno-missing-field-initializers -Wno-microsoft-cast)
```
Malacath-92, line 65: The `-Wshadow-all` should go here

Malacath-92:
Starting variables with an underscore is not a good idea. It's not necessarily disallowed in all contexts here, but sometimes it is and keeping track of all cases where it is and isn't allowed is possibly gonna cause issues some day. And in the case where members previously had the same name it's not a good idea for another reason, namely because those variables most likely refer to separate thing or are used in different context, thus warranting a different name. When we don't know what context that is we should probably do what zapp usually does and name it unknown, e.g. `Qilin::_sound_pos` could instead be `Qilin::sound_pos_unknown`

Mr-Auto:
well, it's not like it's unknown, it probably the same cause with all `ParticleEmitterInfo` that almost all are just named 'particle'

Malacath-92:
Then name it `sound_pos_duplicate`. My point is just that there's almost always a better name than `_sound_pos`

src/game_api/script/usertypes/entities_items_lua.cpp
```diff
@@ -311,8 +311,8 @@ void register_usertypes(sol::state& lua)
         &Spark::rotation_center_x,
         "rotation_center_y",
         &Spark::rotation_center_y,
-        "angle",
-        &Spark::angle,
+        "rotation_angle",
```
Malacath-92, line 314: Is this an API break? Previously when accessing `.angle` would that refer to `Entity.angle` or `Spark.angle`?

Malacath-92 (approved):

src/game_api/script/usertypes/entities_items_lua.cpp
Mr-Auto: just tested with overlunky, it's Spark.angle

to be honest, is soo stupid that VS won't mark this as an error, i can see myself spending hours to find that kind of mistake

src/game_api/entities_monsters.hpp
```diff
@@ -165,7 +165,7 @@ class Mole : public Monster
     uint8_t counter_nonburrowing;
     uint8_t countdown_for_appearing;
     uint8_t unknown_two_while_burrowed;
-    /// 0 non_borrowed, 1 - unknown, 2 - borrowed, 3 - state_change
+    /// 0 - non_borrowed, 1 - unknown, 2 - borrowed, 3 - state_change
```
Malacath-92, line 168: There's also a typo here saying "borrowed" instead of "burrowed"

src/game_api/state_structs.hpp
```diff
@@ -4,6 +4,7 @@
 #include <cstdint>
 
 #define MAX_PLAYERS 4
+using ENT_TYPE = uint32_t; // NoAlias
```
Malacath-92, line 7: This is already defined in entity.hpp, can that not be included here?

Malacath-92 (approved):
You really have some rare endurance to go through this stuff  👏

src/game_api/state_structs.hpp
Mr-Auto: it's not in entity.hpp, it's in lua_backend.hpp, with all the others, it's in here cuz that's covers all the places that don't reference lua_backend.hpp, there is no way to cross include between, exceptionally we could make new header file for all types used across the API, maybe throw some #define there as well and include that everywhere

Malacath-92: It is here https://github.com/spelunky-fyi/overlunky/pull/136/files#diff-4cebcfa05af7753384df6a6b719e6007fa7dd5336bf20937fc1cd86c29911bafR52
Indeed with the one you mentioned there are 3 places this type alias lives. I think it would be better to add them to a separate file, but it's not a bug to have multiple so no hurry to fix that, we can do it some other time so we don't delay merging this more.

Issue #78, opened 2021-05-23
set_timeout is not stoping when instant warp


Mr-Auto:
Probably set_interval as well, i think it should as they are very much used 'per level'

Trixelized:
`set_interval` and `set_timeout` also stop when pausing and entering the options menu

Mr-Auto:
also suggestion: make it so if you return false (or whatever value chosen) in set_interval it will stop it

Dregu:
Fixed and added.

Issue #138, opened 2021-08-16
Cleared callbacks are still executed until the next frame


Malacath-92:
Since we delete cleared callbacks deferred we still execute them once `clear_callback` has been called. We should add guards against that so the behavior is as the user expects while still having the security of deferred deletion.

PR #140, opened 2021-08-17
Clear timers on warp, make level callbacks better

Dregu:
After adding warps the timers were not cleared cause the screen never changes. Also the checks in onlevel were always a bit janky and I think they only worked 99.5%.

I think state.loading should work for both. Does this make more sense?

Also returning false from interval to clear it sounds neat as in #78

### DIFF
--- a/src/game_api/script/lua_backend.cpp
+++ b/src/game_api/script/lua_backend.cpp
@@ -279,8 +279,13 @@ bool LuaBackend::update()
             {
                 if (now >= cb->lastRan + cb->interval)
                 {
-                    handle_function(cb->func);
+                    std::optional<bool> clear_me = handle_function_with_return<bool>(cb->func);
                     cb->lastRan = now;
+                    if (clear_me && clear_me.value() == false)
+                    {
+                        it = level_timers.erase(it);
+                        continue;
+                    }
                 }
                 ++it;
             }
@@ -404,8 +409,13 @@ bool LuaBackend::update()
             {
                 if (now_l >= cb->lastRan + cb->interval)
                 {
-                    handle_function(cb->func);
+                    std::optional<bool> clear_me = handle_function_with_return<bool>(cb->func);
                     cb->lastRan = now_l;
+                    if (clear_me && clear_me.value() == false)
+                    {
+                        it = level_timers.erase(it);
+                        continue;
+                    }
                 }
                 ++it;
             }
--- a/src/game_api/script/lua_backend.cpp
+++ b/src/game_api/script/lua_backend.cpp
@@ -186,11 +186,11 @@ bool LuaBackend::update()
             if (on_screen)
                 on_screen.value()();
         }
-        if (on_frame && g_state->time_level != state.time_level && g_state->screen == 12)
+        if (on_frame && g_state->time_level != state.time_level && g_state->screen == (int)ON::LEVEL)
         {
             on_frame.value()();
         }
-        if (g_state->screen == 11 && state.screen != 11)
+        if (g_state->screen == (int)ON::CAMP && state.screen != (int)ON::CAMP)
         {
             if (on_camp)
                 on_camp.value()();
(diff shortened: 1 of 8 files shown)


src/game_api/script/lua_backend.cpp
```diff
@@ -276,8 +279,13 @@ bool LuaBackend::update()
             {
                 if (now >= cb->lastRan + cb->interval)
                 {
-                    handle_function(cb->func);
+                    std::optional<bool> clear_me = handle_function_with_return<bool>(cb->func);
                     cb->lastRan = now;
+                    if (clear_me && clear_me.value() == false)
```
Malacath-92, line 284: This condition could just be `if (clear_me.value_or(false))`

src/game_api/script/lua_backend.cpp
```diff
@@ -176,10 +176,13 @@ bool LuaBackend::update()
         std::vector<Player*> players = get_players();
         lua["players"] = std::vector<Player*>(players);
 
-        if (g_state->screen != state.screen && g_state->screen_last != 5)
+        if (g_state->loading == 1 && g_state->loading != state.loading)
```
Malacath-92, line 179: Don't we want the `screen != options && screen_last != options` check here?

src/game_api/script/lua_backend.cpp
```diff
@@ -176,10 +176,13 @@ bool LuaBackend::update()
...
         {
             level_timers.clear();
             script_input.clear();
+        }
+        if (g_state->screen != state.screen && g_state->screen_last != 5)
```
Malacath-92, line 184: Instead here, why do we only check moving from options and not moving to options? Or check neither. As it's only a callback users might be interested in moving to the options screen.

src/game_api/script/lua_backend.cpp
```diff
@@ -355,8 +363,7 @@ bool LuaBackend::update()
                 }
                 case ON::START:
                 {
-                    if (g_state->screen == (int)ON::LEVEL && g_state->level_count == 0 && !players.empty() &&
-                        state.player != players.at(0))
+                    if (g_state->screen == (int)ON::LEVEL && g_state->level_count == 0 && state.loading != g_state->loading && g_state->loading == 3)
```
Malacath-92, line 366: This one also runs when coming back from options right now. So I guess it needs a `last_screen != options` or `state.time_level > g_state->time_level`

src/game_api/script/lua_backend.cpp
```diff
@@ -311,12 +319,12 @@ bool LuaBackend::update()
...
         for (auto& [id, callback] : callbacks)
         {
-            if ((ON)g_state->screen == callback.screen && g_state->screen != state.screen && g_state->screen_last != 5) // game screens
+            if ((ON)g_state->screen == callback.screen && g_state->screen != state.screen && g_state->screen_last != (int)ON::OPTIONS) // game screens
             {
                 handle_function(callback.func);
                 callback.lastRan = now;
             }
-            else if (callback.screen == ON::LEVEL && g_state->screen == (int)ON::LEVEL && g_state->screen_last != (int)ON::OPTIONS && !players.empty() && (state.player != players.at(0) || ((g_state->quest_flags & 1) == 0 && state.reset > 0)))
+            else if (callback.screen == ON::LEVEL && g_state->screen == (int)ON::LEVEL && state.loading != g_state->loading && g_state->loading == 3)
```
Malacath-92, line 327: This actually also still runs when coming back from options

src/game_api/script/lua_backend.cpp
Dregu: My brain really went backwards here didn't it if I put the options check in the only place it wasn't needed.

Dregu:
I guess my logic was relying on that visiting options doesn't use state.loading but leaving to main menu somehow does.

Malacath-92:
Added a fix for #138

Malacath-92 (approved):

PR #144, opened 2021-09-01
Docfix2

Mr-Auto:
~~Currently small issue with ImGui and auto doc~~


src/game_api/state_structs.hpp
```diff
@@ -16,10 +16,18 @@ struct LightParams
 
 struct Illumination
 {
-    LightParams light1;
-    LightParams light2;
-    LightParams light3;
-    LightParams light4;
+    union
+    {
+        std::array<LightParams, 4> light;
```
Malacath-92, line 21: Should probably be `lights` plural

Malacath-92: And should probably also document that it's the same as [light1, light2...] since you exposed it to Lua

src/game_api/rpc.cpp
```diff
@@ -423,7 +423,7 @@ std::vector<uint32_t> get_entities_at(ENT_TYPE entity_type, uint32_t mask, float
             {
                 auto [ix, iy] = item->position();
                 float distance = sqrt(pow(x - ix, 2.0f) + pow(y - iy, 2.0f));
-                if (((item->type->search_flags & mask) > 0 || mask == 0) && (item->type->id == entity_type || entity_type == 0) && distance < radius)
+                if (((item->type->search_flags & mask) > 0 || mask == 0) && distance < radius && (entity_types[0] == 0 || std::find(entity_types.begin(), entity_types.end(), item->type->id) != entity_types.end()))
```
Malacath-92, line 426: This sort of API should probably handle receiving empty `entity_types` the same way it handles `{0}`. So we can add a utility to check for a match and defer all these functions to that.

src/injected/ui.cpp
```diff
@@ -4153,10 +4153,18 @@ void render_entity_props()
         }
         for (int i = 0; i < 6; i++)
         {
-            ImGui::CheckboxFlags(button_flags[i], &g_entity->buttons, int_pow(2, i));
+            int buttons = g_entity->buttons;
+            ImGui::CheckboxFlags(button_flags[i], &buttons, int_pow(2, i));
             if (i < 5)
                 ImGui::SameLine(region.x / 6 * (i + 1));
         }
+        //for (int i = 0; i < 6; i++)
```
Malacath-92, line 4161: Stale code, should this be deleted?

PR #145, opened 2021-09-01
Shiny level gen

Malacath-92:
- Fixed a bug that ghost tile codes were always defaulting to sad AI
- Added ability to create and get short tile codes (those are the char-based tilecodes that .lvl files declare)
- Added hook right before a room, made up of short tile codes, handles all it's tiles and spawns entities
    - add a backlayer if not present
    - query and modify short tile codes in front and back layer
- Added new room template type for machine rooms
    - those can have their sizes set via `set_room_template_size`
    - to place one in a level you have to call `mark_as_machine_room_origin` on the top left room
- Added floor spreading spawn type flag
- Fixed an issue with erroneously overlapping flags (LEVEL_GEN_GENERAL should only be set when no other LEVEL_GEN flag is set)
- Added an API for modifying rooms right before their entities are spawned
- Added an API for querying and modifying short tile codes
- Added a function to check whether a room is flipped


src/game_api/entities_monsters.hpp
```diff
@@ -94,6 +94,14 @@ class NPC : public Monster
     uint32_t padding3;
 };
 
+enum class GHOST_BEHAVIOR : uint8_t
+{
+    ANGRY = 0,
+    SURPRISED = 1,
+    SAD = 3,
```
zappatic, line 101: This doesn't match the original comment, I believe it has to be 0, 1, 2, 3

Malacath-92: Oh yes, this is a typo of writing 3 instead of 2, then just adding one for the last  😅

src/game_api/level_api.cpp
```diff
@@ -105,13 +112,13 @@ std::array g_community_tile_codes{
...
-    CommunityTileCode{"ghost", "ENT_TYPE_MONS_GHOST"},
-    CommunityTileCode{"ghost_med_sad", "ENT_TYPE_MONS_GHOST_MEDIUM_SAD"},
-    CommunityTileCode{"ghost_med_happy", "ENT_TYPE_MONS_GHOST_MEDIUM_HAPPY"},
-    CommunityTileCode{"ghost_small_angry", "ENT_TYPE_MONS_GHOST_SMALL_ANGRY"},
-    CommunityTileCode{"ghost_small_sad", "ENT_TYPE_MONS_GHOST_SMALL_SAD"},
-    CommunityTileCode{"ghost_small_surprised", "ENT_TYPE_MONS_GHOST_SMALL_SURPRISED"},
-    CommunityTileCode{"ghost_small_happy", "ENT_TYPE_MONS_GHOST_SMALL_HAPPY"},
+    CommunityTileCode{"ghost", "ENT_TYPE_MONS_GHOST", g_spawn_ghost<GHOST_BEHAVIOR::SAD>},
+    CommunityTileCode{"ghost_med_sad", "ENT_TYPE_MONS_GHOST_MEDIUM_SAD", g_spawn_ghost<GHOST_BEHAVIOR::SAD>},
+    CommunityTileCode{"ghost_med_happy", "ENT_TYPE_MONS_GHOST_MEDIUM_HAPPY", g_spawn_ghost<GHOST_BEHAVIOR::HAPPY>},
```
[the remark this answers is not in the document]
zappatic: I think you have to look at it as three different sets of states (ghost:0 ; medium: 0,1; small: 0, 1, 2, 3)
The original comment named the states in context of the small ghost.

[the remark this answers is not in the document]
Malacath-92: I see. That's still really weird since I'd expect that all ghosts use the same statemachine but now that implies that the state machine code either is separate or that it has to take into account the type id. Either way surprising to me. I'll fix it later

src/game_api/script/usertypes/level_lua.hpp
```diff
@@ -1,13 +1,16 @@
...
 #include <string>
 #include <vector>
 
 #include <sol/forward.hpp>
 
-using ROOM_TEMPLATE = uint16_t;     // NoAlias
-using PROCEDURAL_CHANCE = uint32_t; // NoAlias
-using ROOM_TEMPLATE_TYPE = int32_t; // NoAlias
+using LAYER = int; // NoAlias
+constexpr LAYER LAYER_FRONT = 0;
```
Mr-Auto, line 11: I think this should be uint8_t not LAYER, as LAYER was solely made to be used with enum_to_layer function, first, then you check for the actual layer with value it returns (except LAYER_BOTH). Ideally i was thinking to make it throw warning/error if you try to compare LAYER with int or any other variable that is not LAYER

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -31,10 +35,151 @@ void PostRoomGenerationContext::set_num_extra_spawns(std::uint32_t extra_spawn_i
...
+                }
+            }
+        }
+        return positions;
+    }
+    return {};
+}
+bool PreHandleRoomTilesContext::replace_short_tile_code(LAYER layer, SHORT_TILE_CODE short_tile_code, SHORT_TILE_CODE replacement_short_tile_code)
+{
+    layer = layer < 0 ? 0 : layer;
```
Mr-Auto, line 119: missing enum_to_layer, unless this can't be executed after players are spawned in?

Malacath-92: Happens during room spawning so naturally the player isn't spawned. I should add that to the docs if it's not already there.

src/game_api/script/usertypes/level_lua.hpp
Malacath-92: It's just a workaround for not having a LAYER enum on the c++ side. Used to compare values coming in from the API so there's no value in narrowing the type. Ideally we should have the enum on the C++ side

Mr-Auto: i will add that in my pr, then we can fix which ever pr comes first

src/game_api/flags.hpp
```diff
@@ -85,8 +85,8 @@ const char* more_flags[]{
     "9: ",
     "10: ",
     "11: Swimming",
-    "12: ",
-    "13: ",
+    "12: Hit ground (bomb)",
+    "13: Hit wall (bomb)",
```
Mr-Auto, line 89: This is probably for all thowable entities, not only bomb. Never got time to test it fully

PR #152, opened 2021-10-15
More API stuff

zappatic:


src/game_api/rpc.cpp
```diff
@@ -985,6 +985,26 @@ void drop(uint32_t who_uid, uint32_t what_uid)
     }
 }
 
+void unequip(uint32_t who_uid, uint32_t what_type)
```
Mr-Auto, line 988: Do we need a type? you can't have more then one anyway

zappatic: I would prefer the flat function like drop and pick_up etc. If you do it in powerupcapable, then the modder has to get hold of that entity anyway, and that defeats the purpose of the original simplification you suggested of leaving out the type.

As to what backpack you are wearing: I thought there could be a conflict if you were wearing a jetpack, and picking up a powerpack, both types would be in items and I wouldn't know which was the worn one. But, you can't do that in the game anyway, if you wear a jetpack and pick up a powerpack, you equip it instead. The only way there are two backpacks in items is if you wear and pick up the same thing.
I think this way we can determine what the player is wearing at all times.

Mr-Auto: `get_entity(who_uid):unequip()` vs `unequip(who_uid)` 
adding more functions to types feels more natural for me, keeps all the functions in right place instead of some randomly scattered in the doc. But no big deal.

Dregu: > `get_entity(who_uid):unequip()` vs `unequip(who_uid)` adding more functions to types feels more natural for me, keeps all the functions in right place instead of some randomly scattered in the doc. But no big deal.

By all means add both variants for all functions that concern a single entity/uid. They are currently scattered 50/50 as globals and methods.

Dregu:
When I enable player godmode the HH always get gibbed when they should leave a corpse.

Dregu:
It's not a huge issue, but it's not what you'd expect so people might keep reporting it if you don't fix it :)

PR #153, opened 2021-10-17
Add functions around strings

Mr-Auto:
Changing strings in stringsdb/strings table, adding custom strings etc.


src/game_api/strings.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include "detours.h"
+#include <cassert>
+
+static STRINGID wrong_stringid = 0;
+std::map<STRINGID, std::u16string> custom_strings;
+std::map<uint32_t, STRINGID> string_hashes = {
+    {0xd41f49aa, 0},
+    {0xcd92c15d, 1}
+    // TODO: make script to dump all automatically
```
Malacath-92, line 13: You can easily get this stuff dumped by modlunky or playlunky.

src/game_api/strings.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+const char16_t** get_strings_table()
+{
+    static const char16_t** string_table = (const char16_t**)get_address("string_table");
+    return string_table;
+}
+
+STRINGID hash_to_stringid(uint32_t hash)
+{
+    if (string_hashes.find(hash) != string_hashes.end())
```
Malacath-92, line 63: `auto it = string_hashes.find(hash);` and then `return it->second` is a slight optimization here

src/game_api/strings.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        game_free((void*)*old_string);
+        *old_string = new_string;
+    }
+}
+
+STRINGID add_string(const std::u16string& str) // future idea: add more strings variants for all languages?
+{
+    STRINGID new_id = wrong_stringid + (STRINGID)custom_strings.size() + 1;
+    custom_strings.insert(std::pair<STRINGID, std::u16string>(new_id, str));
```
Malacath-92, line 119: Take the parameter by value `std::u16string str` and then `custom_strings[new_id] = std::move(str);` will avoid extra allocations.

src/game_api/strings.cpp
```diff
@@ -0,0 +1,125 @@
...
+        game_free((void*)*old_string);
+        *old_string = new_string;
+    }
+}
+
+STRINGID add_string(const std::u16string str) // future idea: add more strings variants for all languages?
+{
+    STRINGID new_id = wrong_stringid + (STRINGID)custom_strings.size() + 1;
+    custom_strings.insert(std::pair<STRINGID, std::u16string>(new_id, str));
+    custom_strings[new_id] = std::move(str);
```
Malacath-92, line 123: This didn't work out as I meant  :D
You can't have the parameter as `const` since that prevents to move it.
You also don't want both the insert and the line that moves.

Mr-Auto: got it, i thought i replaced that line, not added new xd

PR #163, opened 2021-10-29
Add the rest of entities from plugin

Mr-Auto:


Dregu:
I would probably swap flags and amount in spawn_liquid, and make flags and separation optional for ease of use.

Mr-Auto:
normally flags are always 0 normally, and i feel like making mess with all the optional with sol, so i rather just add comment "set flags to 0 to ignore"

PR #165, opened 2021-11-02
API enhancements

zappatic:


zappatic:
For now, yes, they are just allocated, never deallocated. Still have to look into that.

PR #167, opened 2021-11-05
Add enter_door, change_sunchallenge_spawn, change_diceshop_prizes functions, custom ENT_TYPE's and more

Mr-Auto:


src/game_api/rpc.cpp
```diff
@@ -1657,3 +1657,20 @@ bool entity_type_check(const std::vector<ENT_TYPE>& types_array, const ENT_TYPE
 
     return false;
 }
+
+void enter_door(Entity* player, Entity* door)
```
Dregu, line 1661: Convert to uids, we don't really use Entities as arguments on the user side.

Dregu:
Also I still don't see the point of any of this, cause you could always do this through state variables without having to maintain patterns for a bunch of functions. See `warp` for example. It used to use a game function, but the pattern broke/it was inlined, so I just replicated the simple functionality. I assume these functions are ultimately just changing state variables as well... Adding these might just make updates harder for us without actually adding anything to the api, other than simplifying the task for the modder. Replicating the functionality as helper functions might be better for everyone...

Mr-Auto:
Changing stuff in state does not call this function, and the function itself does more stuff, not sure how relevant

It's not hard to maintain and if it ever brakes/gets inlined we can just write it with changing values in state. Also i like doing stuff as the game does more then changing some values.

Also having a function is easier for people to understand, than look for/read thru some discord posts to find a way to call transition

Mr-Auto:
ok, i don't get this. When i build this locally, state_lua in not even in top 10 heaviest to build, i even rebuild the whole thing and still had no problem when building the lua stuff :thinking:

Mr-Auto:
I'll guess the get_metadata in the doc is bug in generator?

src/game_api/render_api.hpp
```diff
@@ -68,15 +68,24 @@ struct RenderInfo
     float x_dupe2;
     float y_dupe2;
     uint32_t unknown11;
-    uint32_t unknown12;
-    uint32_t unknown13;
+    uint8_t unknown12a;
+    uint8_t unknown12b;
+    bool unknown12c;
+    bool unknown12d;
+    bool unknown13a;
```
Dregu, line 75: unknown13a is true when the entity is culled out, not rendered at all, and get_render_position should revert to get_position

Mr-Auto: well, i think it shouldn't return position at all in this situation, like, it's not rendered
also the function shouldn't return 0.0 when it's wrong entity as that's valid position, but i guess it has to stay to not break compatibility

Mr-Auto: i could potentially write the same stuff as i did for the draw olmec hitbox, so loop thru items until on that is drawn is found and return: it's rendering - offset

src/game_api/entity.cpp
```diff
@@ -244,22 +244,14 @@ std::pair<float, float> Entity::position_self() const
 
 std::pair<float, float> Entity::position_render() const
 {
-    if (overlay != nullptr)
-    {
-        auto [x_pos, y_pos] = position_self();
-        auto [rx_pos, ry_pos] = overlay->position_render();
-        return {rx_pos + x_pos, ry_pos + y_pos};
-    }
-    if (rendering_info == nullptr)
```
Malacath-92, line 253: Why drop this verification?

src/game_api/entity.cpp
```diff
@@ -244,22 +244,14 @@ std::pair<float, float> Entity::position_self() const
 
 std::pair<float, float> Entity::position_render() const
 {
-    if (overlay != nullptr)
```
Malacath-92, line 247: And this is not necessary?

src/game_api/entity.cpp
```diff
@@ -413,8 +405,8 @@ std::tuple<float, float, uint8_t> get_position(uint32_t uid)
 std::tuple<float, float, uint8_t> get_render_position(uint32_t uid)
 {
     Entity* ent = get_entity_ptr(uid);
-    if (ent)
-        return std::make_tuple(ent->position_render().first, ent->position_render().second, ent->layer);
+    if (ent && !ent->rendering_info->stop_render)
```
Malacath-92, line 408: Why? Assuming `stop_render` means that the entity shall not be rendered. That does seem to imply that we should not return the given position here. Is there a good motivation for this?

src/game_api/entity.cpp
```diff
@@ -621,15 +619,30 @@ void Entity::set_on_damage(std::uint32_t reserved_callback_id, std::function<boo
 
 bool Entity::is_movable()
 {
+    static ENT_TYPE first_logical = to_id("ENT_TYPE_LOGICAL_CONSTELLATION");
```
Malacath-92, line 622: Should be `static const` for correctness.

src/game_api/entity.hpp
```diff
@@ -129,6 +129,7 @@ struct EntityDB
     int32_t sound_killed_by_other;
     float field_a8;
     int32_t field_AC;
+    /// Array of [Animation](#Animation) types, had to use `pairs` loop to loop thru
```
Malacath-92, line 132: `had to use `pairs` loop to loop thru` seems like redundant information, you always have to use `pairs` or `ipairs` to loop and this being a map makes it clear that `ipairs` will not suffice

src/game_api/entity.hpp
```diff
@@ -361,29 +363,23 @@ struct Inventory
 {
     uint32_t money;
     uint8_t bombs;
+    /// To access/edit anything below use `ON.PRE_LEVEL_GENERATION`, those are used to transfer information to new player entity
     uint8_t ropes;
-    /// Used in level transition to transfer to new player entity, is wrong during the level
```
Malacath-92, line 365: Maybe keep a `see "ropes" for correct usage`

src/game_api/flags.hpp
```diff
@@ -292,30 +293,15 @@ const char* illumination_flags[]{
...
-    "32: ",
+    // type_flags 1-8:
+    "9: Follow camera",
+    "10: Follow entity",
+    "11: Static, Rectangle, Full brightness",
+    "12: unused",
+    "13: unused",
+    "14: unused",
+    "15: unused",
+    "16: unused",
```
Malacath-92, line 304: Everything past the 16th flag was wrong? Or was it moved to another flags field?

src/game_api/layer.hpp
```diff
@@ -1,28 +1,96 @@
...
 #include "entity.hpp"
 
+struct Entities
+{
+    Entity** ent_list;
+    uint32_t* uids;
+    uint32_t cap;
+    uint32_t size;
+
+    constexpr Entity** begin()
```
Malacath-92, line 11: For even easier usage we could return a `struct EntitiesIterable;` that has the `begin` and `end` functions defined as this. And then do the same with `UidsIterable` doing the same so you can iterate either entities or uids in a range-based for-loop.

src/game_api/memory.hpp
```diff
@@ -7,6 +7,80 @@
 #include "search.hpp"
 
 using namespace std::string_literals;
+class ExecutableMemory
```
Malacath-92, line 10: Is there a good reason to move this? If not I'd prefer to put it back to where it was to reduce the size of the diff.

src/game_api/memory.hpp
```diff
@@ -7,6 +7,80 @@
...
+  private:
+    struct deleter_t
+    {
+        void operator()(std::byte* mem) const;
+    };
+    using storage_t = std::unique_ptr<std::byte, deleter_t>;
+    storage_t code;
+};
+
+struct Memory
```
Malacath-92, line 43: Same here.

src/game_api/memory.hpp
```diff
@@ -59,10 +133,61 @@ void write_mem_prot(size_t addr, std::string payload, bool prot)
...
+struct ReversibleMemory
+{
+    size_t address;
+    std::vector<char> old_data;
+    bool prot_used;
+};
+
+std::unordered_map<std::string, std::vector<ReversibleMemory>> original_memory;
+
+[[maybe_unused]] void write_mem_reversible(std::string name, size_t addr, std::string_view payload, bool prot)
```
Malacath-92, line 146: `write_mem_recoverable` seems better here.

src/game_api/memory.hpp
```diff
@@ -59,10 +133,61 @@ void write_mem_prot(size_t addr, std::string payload, bool prot)
...
+        if (new_addr)
+        {
+            std::vector<char> old_data((char*)addr, (char*)addr + payload.size());
+            original_memory[name].push_back(ReversibleMemory{addr, old_data, prot});
+        }
+    }
+    write_mem_prot(addr, payload, prot);
+}
+template <class T>
+requires(std::is_trivially_copyable_v<T> && !std::is_same_v<T, std::string_view>) void write_mem_reversible(std::string name, size_t addr, const T& payload, bool prot)
```
Malacath-92, line 173: Can you move the concept to a new line and wrap the whole function in `// clang-format off` and `// clang-format on` please?
This is kind of unreadable  :(

src/game_api/memory.hpp
```diff
@@ -95,82 +220,28 @@ FunT* vtable_find(T* obj, size_t index)
         return static_cast<FunT*>(nullptr);
     return reinterpret_cast<FunT*>(&ptr[0][index]);
 }
-}; // namespace
 
-class ExecutableMemory
```
Malacath-92, line 100: Diff is very difficult to read here, I will get back to it when you moved the two classes back  :P

src/game_api/rpc.cpp
```diff
@@ -499,11 +438,11 @@ Entity* get_entity_ptr(uint32_t uid)
 
 ENT_TYPE get_entity_type(uint32_t uid)
 {
-    auto state = State::get();
-    auto p = state.find(uid);
-    if (p == nullptr || IsBadWritePtr(p, 0x178))
-        return UINT32_MAX;
-    return p->type->id;
+    auto p = get_entity_ptr(uid);
```
Malacath-92, line 441: We don't like var names like `p`  ;)

src/game_api/rpc.cpp
```diff
@@ -565,71 +504,65 @@ std::vector<uint32_t> filter_entities(std::vector<uint32_t> entities, std::funct
...
 std::vector<uint32_t> get_entities()
 {
-    return get_entities_by({}, 0, LAYER::BOTH);
+    return get_entities_by(0, 0, LAYER::BOTH);
 }
 
 std::vector<uint32_t> get_entities_by_layer(LAYER layer)
 {
-    return get_entities_by({}, 0, layer);
+    return get_entities_by(0, 0, layer);
```
Malacath-92, line 512: This might be slightly more readable, but it allocates memory at no additional functionality. So please revert this, unless I missed some subtle details.

src/game_api/rpc.cpp
```diff
@@ -1765,3 +1708,220 @@ bool entity_type_check(const std::vector<ENT_TYPE>& types_array, const ENT_TYPE
...
+    auto door = get_entity_ptr(door_uid);
+    if (player == nullptr || door == nullptr)
+        return;
+
+    typedef void transition_fun(Entity*, Entity*);
+    static transition_fun* door_entry = (transition_fun*)(addr);
+    door_entry(door, player);
+}
+
+void change_sunchallenge_spawn(std::vector<ENT_TYPE> ent_types)
```
Malacath-92, line 1742: Should this not be plural `change_sunchallange_spawns`?

src/game_api/script/usertypes/entities_chars_lua.cpp
```diff
@@ -34,6 +34,23 @@ void register_usertypes(sol::state& lua)
         &Inventory::money,
         "bombs",
         &Inventory::bombs,
+        "time_of_death",
+        &Inventory::time_of_death,
+        "player_slot",
+        sol::property([](Inventory& i) -> int8_t
```
Malacath-92, line 40: Docs don't handle this well, right? Place in a local variable and move in here to make the docs look nicer.

src/game_api/script/usertypes/gui_lua.cpp
```diff
@@ -94,19 +94,19 @@ void GuiDrawContext::draw_text(float x, float y, float size, std::string text, u
...
-    ImVec2 b = screenify({x2, y2});
+    ImVec2 a = screenify({left, top});
+    ImVec2 b = screenify({right, bottom});
     ImVec2 uva = ImVec2(uvx1, uvy1);
     ImVec2 uvb = ImVec2(uvx2, uvy2);
     backend->draw_list->AddImage(backend->images[image]->texture, a, b, uva, uvb, color);
 };
 void GuiDrawContext::draw_image(IMAGE image, AABB rect, AABB uv_rect, uColor color)
 {
-    draw_image(image, rect.left, rect.bottom, rect.right, rect.top, uv_rect.left, uv_rect.bottom, uv_rect.right, uv_rect.top, color);
```
Malacath-92, line 109: This bottom vs top thing again  :notlikethis:
Are you sure this is right? We broke this once before.

src/game_api/search.cpp
```diff
@@ -1582,6 +1582,67 @@ std::unordered_map<std::string_view, AddressRule> g_address_rules{
...
+            .at_exe(),
+    },
+    {
+        // Set write bp on Movable.poison_tick_timer, get hit by cobra's acid spit
+        "poison_entity"sv,
+        PatternCommandBuffer{}
+            .find_inst("\x48\x8B\x4E\x08\xF6\x41\x51\x04"sv)
+            .at_exe()
+            .function_start(),
+    },
```
Malacath-92, line 1645: Fyi, I didn't check any patterns

src/game_api/spawn_api.cpp
```diff
@@ -74,28 +74,33 @@ void spawn_liquid(ENT_TYPE entity_type, float x, float y)
 
 void spawn_liquid(ENT_TYPE entity_type, float x, float y, float velocityx, float velocityy, uint32_t liquid_flags, uint32_t amount, float blobs_separation = INFINITY)
 {
-    LiquidPhysicsParams* liquid_physics = nullptr;
-    auto state_ptr = State::get().ptr();
-
-    if (entity_type == to_id("ENT_TYPE_LIQUID_WATER"))
+    LiquidTileSpawnData* liquid_physics = nullptr;
```
Malacath-92, line 77: Maybe rename the var here?

src/game_api/spawn_api.cpp
```diff
@@ -74,28 +74,33 @@ void spawn_liquid(ENT_TYPE entity_type, float x, float y)
...
+    auto state_ptr = get_state_ptr();
+    static ENT_TYPE liquid_water = to_id("ENT_TYPE_LIQUID_WATER");
```
Malacath-92, line 79: All of those should be `const`

src/game_api/state.cpp
```diff
@@ -456,3 +456,34 @@ Entity* State::find(uint32_t uid)
         cur_index = (cur_index + 1) & mask;
     }
 }
+
+LiquidPhysicsEngine* State::get_correct_liquid_engine(ENT_TYPE liquid_type)
+{
+    const auto state = ptr();
+    static ENT_TYPE LIQUID_WATER = to_id("ENT_TYPE_LIQUID_WATER"sv);
```
Malacath-92, line 463: All should be `const`

src/game_api/state_structs.hpp
```diff
@@ -737,19 +751,22 @@ struct LiquidPhysicsParams
...
-        struct
-        {
-            LiquidPhysicsParams water_physics;
-            LiquidPhysicsParams coarse_water_physics;
-            LiquidPhysicsParams lava_physics;
-            LiquidPhysicsParams coarse_lava_physics;
-            LiquidPhysicsParams stagnant_lava_physics;
-        };
-    };
+    size_t unknown1; // MysteryLiquidPointer1 in plugin
```
Malacath-92, line 754: Those three vals could be in a struct and then we can put this stuff back into the same union, just putting that struct instead of LiquidPhysicsParams.

Malacath-92 (changes requested):

src/game_api/entities_floors.hpp
```diff
@@ -63,9 +63,7 @@ class Door : public Floor
 {
   public:
     uint8_t counter; // counts down as you go thru it
-    int8_t unused1;
-    int16_t unused2;
-    int32_t unused3;
+    int8_t unused1[7];
```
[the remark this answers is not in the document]
Mr-Auto: what do you mean? it should be 7 bytes, 
unused1 = 1 byte, unused2 = 2 bytes, unused3 = 4 bytes
is my math not correct?
also it's broken in current version but this fixes it

src/game_api/entity.hpp
Mr-Auto: wdym?

src/game_api/flags.hpp
Mr-Auto: those are bool's not flags, so i moved them to separate variable, also made sure that it's not breaking anything with struct and union stuff

src/game_api/memory.hpp
Mr-Auto: needed to have this above the functions as they use get function from them, and functions needs to be in the `namespace
{` thing... for whatever reason, otherwise it just throws tons of errors
i actually hate this file, i prefer to have the usual header/source as two separate files, but i didn't know if this was made this way on purpose or not

src/game_api/rpc.cpp
Mr-Auto: don't point at me, i didn't make it originally xd

src/game_api/script/usertypes/gui_lua.cpp
Mr-Auto: this makes sense if compared to all the _rect functions now, why would this be different?
also i think no one is using AABB xd

src/game_api/state_structs.hpp
Mr-Auto: yeah, but why? is there a need to do it that way?

src/game_api/entity.cpp
Mr-Auto: there is negation there "!" so it will return position if the `stop_render` is false

src/game_api/character_def.cpp
```diff
@@ -118,36 +117,15 @@ bool get_character_gender(std::uint32_t character_index)
...
-    const char16_t** short_name = string_table + char_def.short_name_string_id;
-
-    const auto data_size = name.size() * sizeof(char16_t);
-    char16_t* new_short_name = (char16_t*)game_malloc(data_size + sizeof(char16_t));
-    new_short_name[name.size()] = u'\0';
-    memcpy(new_short_name, name.data(), data_size);
-
-    game_free((void*)*short_name);
-    *short_name = new_short_name;
+    change_string(char_def.short_name_string_id, name.data());
```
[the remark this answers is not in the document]
Malacath-92: I see it actually takes a string view, so just don't call ´.data()´ here and you're good.

Malacath-92 (changes requested):

src/game_api/entity.hpp
Mr-Auto: that's not a type, it's typedef

src/game_api/entity.hpp
Mr-Auto: all of those comments say the same thing, that's way i made it into one comment that's say "hey, from this point onward all the vars are level transition use only", and it's added "to the rope" var so it displays under it, so anything below the rope

src/game_api/entity.cpp
Mr-Auto: Render stuff is not exposed, so there is no way telling if the returned value from this function is valid or not

src/game_api/entity.cpp
Mr-Auto: well you could use that argument for any code, like we have tons of functions that access stuff in state. It's a little over protective to check if every single pointer

src/game_api/entity.hpp
Malacath-92: I don't care whether that's a type or a typedef, the auto-docs is just copying strings. So I would assume it says `AnimationMap` so the docs are clear in communicating that it is a map, hence iteration with `pairs` is implied. And I just checked the docs, they do indeed say that. That was my only point, the docs you added are redundant, but they don't hurt so if you have strong feelings about it just keep them.

src/game_api/entity.hpp
Malacath-92: Here my point is also about clarity of docs. Putting a comment on one thing is bound to confuse people. If I go into the docs and just check `Inventory::health` then I don't know what the usage of it is because I would have to look at `Inventory::ropes`. To clarify that you either duplicate the comment on `Inventory::ropes` for each relevant member or you mention that same restrictions apply for each member. Having duplication in docs is not bad in a case like this, removing the duplication is harmful instead.
This one I have strong feelings about  ;)

PR #170, opened 2021-11-08
Dont cache heap base

Malacath-92:
Maybe fixes the wrong-theme bug that is triggered from playing online


Malacath-92:
Oops, I accidentally put some other changes in this PR. Nothing that would have functional changes though. Main point is in thread_utils.cpp

zappatic (approved):

PR #175, opened 2021-12-24
add more patterns to replace_drop, fix some minor stuff

Mr-Auto:


Mr-Auto:
i take ideas for better name than `modify_ankh_health_gain`

src/game_api/drops.hpp
```diff
@@ -14,7 +14,6 @@ struct DropEntry
     uint32_t vtable_rel_offset;
     uint8_t value_offset = 0;      // the offset of the value to be replaced within the pattern
     uint8_t vtable_occurrence = 1; // when a value occurs more than once in the same virtual table function, choose how many times to replace
```
Mr-Auto, line 16: can i remove `vtable_occurrence` ?
it's not used and makes the code for replace_drop awful

src/game_api/entities_floors.hpp
```diff
@@ -251,6 +269,12 @@ class BigSpearTrap : public Floor
     uint8_t active; /*unsure*/ // forced to 1
     /// setting the left part to 0 or right part to 1 destroys the trap
     bool left_part;
+    /// The uid must be movable entity for ownership transfers, has to be called on the left part of the trap,
+    /// direction: 1 = left, 2 = right
```
Malacath-92, line 273: Maybe a `bool left` instead?

src/game_api/entities_floors.hpp
```diff
@@ -365,3 +389,13 @@ class PoleDeco : public Floor
     int32_t deco_up;
     int32_t deco_down;
 };
+
+class JungleSpearTrap : public Floor
+{
+  public:
+    /// The uid must be movable entity for ownership transfers, direction: 1 = left, 2 = right, 3 = up, 4 = down
```
Malacath-92, line 396: Maybe an enum instead?

src/game_api/layer.hpp
```diff
@@ -1,13 +1,19 @@
...
+struct EntityList
 {
     Entity** ent_list;
     uint32_t* uids;
     uint32_t cap;
     uint32_t size;
 
+    constexpr bool empty()
+    {
+        return !!size;
```
Malacath-92, line 14: Please don't start using `!!` but be explicit `size == 0`

src/game_api/layer.hpp
```diff
@@ -26,6 +44,15 @@ struct Entities
         return uids + size;
     };
 
+    constexpr bool contains(Entity* ent)
+    {
+        return (std::find(begin(), end(), ent) != end());
```
Malacath-92, line 49: Unnecessary parens around the whole expression.

src/game_api/layer.hpp
```diff
@@ -26,6 +44,15 @@ struct Entities
         return uids + size;
     };
 
+    constexpr bool contains(Entity* ent)
+    {
+        return (std::find(begin(), end(), ent) != end());
+    };
+    constexpr bool contains(uint32_t uid)
+    {
+        return (std::find(uid_begin(), uid_end(), uid) != uid_end());
```
Malacath-92, line 53: Ditto

src/game_api/level_api.cpp
```diff
@@ -462,6 +462,7 @@ std::array g_community_tile_codes{
             };
             g_attachee_requiring_entities.push_back({{{x, y - 1}}, do_spawn});
         }},
+    CommunityTileCode{"boombox", "ENT_TYPE_ITEM_BOOMBOX"},
```
Malacath-92, line 465: Is this in the community doc?

src/game_api/drops.hpp
Malacath-92: How does it make that code worse exactly?

Malacath-92 (changes requested):

src/game_api/level_api.cpp
Mr-Auto: yeah, we were comparing the ML and OL tile codes and this one was missing, it's in the first wave in the doc

src/game_api/drops.hpp
Mr-Auto: it's just extra code, looks less clear in my opinion

src/game_api/layer.hpp
Mr-Auto: This was just copied straight from the `Vector` type

src/game_api/drops.hpp
Malacath-92: But it's more accurate in case this scenario does eventually happen.

src/game_api/layer.hpp
Malacath-92: Then it's a good opportunity to make it nicer  :)

src/game_api/drops.hpp
Mr-Auto: Isn't that a rule of programming? don't solve problems that don't exist yet

src/game_api/entities_floors.hpp
Mr-Auto: If i could make it also for something else, but making global enum just for this one function feels wasteful

src/game_api/drops.hpp
zappatic: Just FYI: This was obviously added because it was needed at the time. The 1.21.0c (i believe) release invalidated 90% of the patterns, and when those were re-implemented, this was simply not used any more in any of the patterns. While I agree with your rule, the past is also a thing. 
Also, why are the `SACRIFICE_*` drops added when there were similar ones as `ALTAR_*` ? Aren't those duplicates?

Mr-Auto: @zappatic  The `ALTAR_*` ones replace what you get, for example: `ALTAR_DICE_ROPEPILE` replaces rope pile
`SACRIFICE_*` replace what you sacrifice (what you put on the altar)

PR #176, opened 2022-01-07
qilin.lua: Added script to practise the qilin flight challenge

mattlennon3:
As titled, when enabled this script will start any singleplayer game at Tiamat, mount you on the Qilin and off you go :)

If you get stunned or hit by one of the beams you can just use the `DOOR/BUY` keybind to reset your position to just below the jump challenge. 

I made this because we've only just started getting to this point consistently in our multiplayer runs and I was not confident at all doing the challenge! I had to make this so we're less likely to fail and waste 30+ mins and a good run.

### DIFF
--- a/examples/qilin.lua
+++ b/examples/qilin.lua
@@ -1,0 +1,62 @@
+meta= {
+    name = "Qilin Practise",
+    description = "Start with a Qilin at Tiamat to practise flying (press the DOOR/BUY button to reset)",
+    author = "Tuquar"
+}
+
+local qilin_uid = nil
+local start_x = 17.5
+local start_y = 88
+
+local function go_to_tiamat()
+	warp(1, 1, 14)
+end
+
+local function mount_new_qilin()
+	-- Spawn qilin and tame it
+	x, y, layer = get_position(players[1].uid)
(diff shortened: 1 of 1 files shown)


gmjosack:
Why not just upload it as a mod to spelunky.fyi? Is there something novel here that is should be included with the overlunky distribution?

mattlennon3:
@gmjosack I was not aware of spelunky.fyi, I only discovered this project until about 2 days ago. I have avoided most media aside from the wiki for fear of spoilers. A friend sent me a link to this and I thought it was pretty cool and really good timing to help us beat this part of the game.

I just enjoyed some of the scripts and figured I'd make my own to add to the collection.

gmjosack:
The scripts here are mostly meant as examples of features in the lua api. The main method of mod distribution happens via spelunky.fyi and can be played w/ playlunky v/ the modlunky launcher. We already have a couple qilin practice mods uploaded to spelunky.fyi by the way. Appreciate the contribution but I don't think we want to use Overlunky as a large bundle of user created mods.

mattlennon3:
Fair enough. Perhaps you can update the readme to link to that modding site.

PR #178, opened 2022-01-15
CustomTheme can override all sorts of level gen logic

Dregu:


src/game_api/level_api.hpp
```diff
@@ -462,3 +466,280 @@ void add_next_levels(std::vector<std::string> next_levels);
 
 int8_t get_co_subtheme();
 void force_co_subtheme(int8_t subtheme);
+
+class CustomTheme : public ThemeInfo
+{
+  public:
+    uint8_t theme = 100;
```
Dregu, line 474: This list is a clusterfuck, but the idea is you can disable individual aspects of the level gen or get them from a different base theme

Dregu:
I didn't really edit most of those files btw, I just ran clang-format on clang 13 and that stupid shit formats everything a bit differently with the same config.

docs/script-api.md
```diff
@@ -1719,7 +1727,7 @@ airtime = falling_timer
 - [`nil set_cursed(bool b)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_cursed) &Movable::set_cursed
 - [`nil drop(Entity entity_to_drop)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=drop) &Movable::drop
 - [`nil pick_up(Entity entity_to_pick_up)`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=pick_up) &Movable::pick_up
-- [`bool can_jump()`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=can_jump) &Movable::can_jump
+- [`can_jump`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=can_jump) &Movable::can_jump
```
Mr-Auto, line 1730: i don't understand auto doc anymore

src/game_api/level_api.hpp
```diff
@@ -169,80 +176,91 @@ class ThemeInfo
...
     // volcana: checks if state.coffin_contents = cocovondiamonds -> chooses one of the four coffins in vlad's castle at random to put her in
     // tidepool: spawns impostor lake, some door and a litwalltorch, ...
     // neobab: assigns the correct animation_frame to all the ushabtis
     // co, arena: forwards to the same virtual of the sub_theme
     // abzu, tiamat: spawns impostor lake
+    // ???
     virtual void post_process_level() = 0;
 
+    // adds theme specific random traps and pushblocks
     // quillback: adds butterflies and snails + configures the door to go to the correct theme/level
```
Malacath-92, line 216: I assume this comment about butterflies is wrong now?

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -189,6 +189,752 @@ LevelGenRoomData& PreHandleRoomTilesContext::get_mutable_room_data()
...
+    {
+    }
+    ~CustomTheme()
+    {
+    }
+    bool get_unknown1()
+    {
+        return unknown1;
+    }
+    void init_flags()
```
Malacath-92, line 390: Maybe do some sort of macro or template here to avoid the amount of duplicate code for each function?

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -189,6 +189,752 @@ LevelGenRoomData& PreHandleRoomTilesContext::get_mutable_room_data()
...
+
+class CustomTheme : public ThemeInfo
+{
+  public:
+    std::string level_file;
+    uint8_t theme;
+    uint8_t base_theme;
+    LuaBackend* backend;
+
+    std::map<THEME_OVERRIDE, ThemeOverride*> overrides;
```
Malacath-92, line 209: unique_ptr?

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -189,6 +189,752 @@ LevelGenRoomData& PreHandleRoomTilesContext::get_mutable_room_data()
...
+    }
+    CustomTheme(uint8_t theme_id_, uint8_t base_theme_)
+        : CustomTheme(theme_id_, base_theme_, true)
+    {
+    }
+    CustomTheme()
+        : CustomTheme(100, 1, true)
+    {
+    }
+    ~CustomTheme()
```
Malacath-92, line 383: The pointers in overrides leak here if I am not missing anything

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -357,6 +1103,297 @@ void register_usertypes(sol::state& lua)
...
+        &ThemeInfo::spawn_extra,
+        "unknown_v51",
+        &ThemeInfo::unknown_v51);
+
+    auto theme_override = sol::overload(
+        static_cast<void (CustomTheme::*)(THEME_OVERRIDE, bool)>(&CustomTheme::override),
+        static_cast<void (CustomTheme::*)(THEME_OVERRIDE, uint8_t)>(&CustomTheme::override),
+        static_cast<void (CustomTheme::*)(THEME_OVERRIDE, sol::function)>(&CustomTheme::override));
+
+    /// Customizable ThemeInfo with ability to override certain theming functions from different themes or write custom functions. Warning: We WILL change these function names, especially the unknown ones, when you figure out what they do.
```
Malacath-92, line 1218: With the warning it's fair to break backwards compat imho  👍

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -357,6 +1103,297 @@ void register_usertypes(sol::state& lua)
...
+
+    /// Force a theme in PRE_LOAD_LEVEL_FILES, POST_ROOM_GENERATION or PRE_LEVEL_GENERATION to change different aspects of the levelgen. You can pass a CustomTheme, ThemeInfo or THEME.
```
Malacath-92, line 1365: Should we not put those on the relevant contexts instead of a global function=

src/game_api/level_api.hpp
Dregu: Maybe, I don't know if the virtuals got shuffled or changed or not tested correctly, but none of the old comments are to be trusted.

src/game_api/script/usertypes/level_lua.cpp
Dregu: Yeah especially the void ones with no special cases could use a template.

src/game_api/script/usertypes/level_lua.cpp
Dregu: Well the thing is I don't know all the relevant contexts and if you haphazardly pick some, nobody will ever know.

Dregu:
I figured making unique_ptrs was probably enough to delete them too. Also *that's why* what I was doing felt stupid, like there has to be an easier way to do this, but I was just copying what we were doing elsewhere 🤷🏻‍♂️ 

I don't really have anything to add for a nightly release, apart from the optimization that can be done later when we know what things even do... And defaults probably could be tweaked at some point.

Malacath-92:
Ofc make unique doesn't work with aggregates... sometimes I wish I could just fork C++ and fix these kinds of unnecessary restrictions...

src/game_api/script/usertypes/level_lua.cpp
Malacath-92: The contexts are `PreLoadLevelFilesContext` but `PostRoomGenerationContext` but `PRE_LEVEL_GENERATION` doesn't have one but adding one should be simple. Personally I think it'd be better to have things that only work in some context on the relevant contexts so we don't get people complaining that they don't work. I can move these if you want can't be arsed yourself, but ofc I won't if you just plain disagree  :D

PR #179, opened 2022-01-20
Update

Mr-Auto:


Malacath-92:
I merged this into [Update2252](https://github.com/spelunky-fyi/overlunky/tree/Update2252) for now

PR #182, opened 2022-01-22
Add tilecode aliases

jaythebusinessgoose:
This lets multiple tilecode names to map to the same tile code.

### DIFF
--- a/src/game_api/level_api.cpp
+++ b/src/game_api/level_api.cpp
@@ -31,6 +31,7 @@ std::uint32_t g_last_community_chance_id;
 std::uint32_t g_current_chance_id;
 
 std::unordered_map<std::uint32_t, std::string_view> g_tile_code_id_to_name;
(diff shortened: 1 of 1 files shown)


src/game_api/level_api.cpp
```diff
@@ -622,6 +649,13 @@ void handle_tile_code(LevelGenSystem* self, std::uint32_t tile_code, std::uint16
         }
```
jaythebusinessgoose, line 649: I'm assuming that this logic here is the pre-tile-code-callback that lua scripts use, so I'm putting all of the alias stuff after this since clients would obviously want the original tile code name to be what they are editing and not some arbitrarily mapped name.

jaythebusinessgoose:
The list of aliases came from https://docs.google.com/spreadsheets/d/1nDBXjhCI-12gXNN2bRHPi5CZHzgrzBvS9cUQb_w_FE4/edit#gid=249428348

plus skeleton_key

PR #187, opened 2022-02-03
Post pr suggestions

Malacath-92:
@Mr-Auto if you can check that I didn't break anything

### DIFF
--- a/src/game_api/rpc.cpp
+++ b/src/game_api/rpc.cpp
@@ -545,7 +545,9 @@ std::vector<uint32_t> get_entities_by_mask(uint32_t mask)
     return get_entities_by({}, mask, LAYER::BOTH);
 }
 
-void foreach_mask(uint32_t mask, Layer* l, std::function<void(EntityList&)> fun)
+template<class FunT>
+requires std::is_invocable_v<FunT, const EntityList&>
+void foreach_mask(uint32_t mask, Layer* l, FunT&& fun)
 {
     if (mask == 0)
     {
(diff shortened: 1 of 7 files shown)


src/game_api/rpc.cpp
```diff
@@ -573,34 +574,39 @@ std::vector<uint32_t> get_entities_by(std::vector<ENT_TYPE> entity_types, uint32
...
             {
                 found.push_back(item->uid);
             }
         }
     };
-    auto insert_all_uids = [&found](EntityList entities)
+    auto insert_all_uids = [&found](const EntityList& entities)
     {
-        found.insert(found.end(), entities.uid_begin(), entities.uid_end());
+        found.reserve(found.size() + entities.size);
```
Mr-Auto, line 589: isn't this redundant? from what i've read, `insert` already does that if you give it a range

src/game_api/rpc.cpp
```diff
@@ -795,12 +801,13 @@ std::vector<uint32_t> entity_get_items_by(uint32_t uid, std::vector<ENT_TYPE> en
         const std::vector<ENT_TYPE> proper_types = get_proper_types(std::move(entity_types));
         if ((!proper_types.size() || !proper_types[0]) && !mask) // all items
         {
-            found.reserve(entity->items.size);
-            found.insert(found.end(), entity->items.uid_begin(), entity->items.uid_end());
+            found.reserve(found.size() + entity->items.size);
```
Mr-Auto, line 804: there is not way that `found` will have any elements at this point is it? the whole `reserve` can be omitted for the same reason as in the first comment (not sure why i put it here)

src/game_api/rpc.cpp
Malacath-92: I looked at the implementation in the MS STL and it doesn't do that. There's also nothing like that mandated by the standard.

Mr-Auto: You sure? http://www.cplusplus.com/reference/vector/vector/insert/#complexity

Malacath-92: Not super sure, but I do not trust cplusplus.com, it has tons of incorrect information all over the place, sometimes literally contradicting the standard. cppreference.com does not name any complexity requirements in terms of allocations. And as I said, the MS STL does not do this either.

Malacath-92: I'll remove the two lines just so we get these changes in.

Mr-Auto: well, it doesn't really change much if you leave or delete them anyway

Dregu:
So is this fix ready then

Malacath-92:
Yes, I forgot to merge after the last commit

PR #189, opened 2022-02-04
Work on virtuals, fix liquids crashing the game when falling outside of a level and other stuff

Mr-Auto:


src/game_api/entities_items.hpp
```diff
@@ -65,14 +75,24 @@ class VladsCape : public Cape
     bool can_double_jump;
 };
 
-class Mattock : public Movable
+class Purchesable : public Movable
```
gmjosack, line 78: Should this be spelled `Purchasable`?

I keep reading it as per-cheese-able

src/game_api/script/usertypes/entities_items_lua.cpp
```diff
@@ -10,6 +10,10 @@ namespace NEntitiesItems
 void register_usertypes(sol::state& lua)
 {
     lua["Entity"]["as_bomb"] = &Entity::as<Bomb>;
+    lua["Entity"]["as_projectile"] = &Entity::as<Projectile>;
+    lua["Entity"]["as_Purchasable"] = &Entity::as<Purchasable>;
+    lua["Entity"]["as_bow"] = &Entity::as<Bow>;
+    lua["Entity"]["as_dummyPurchasableentity"] = &Entity::as<DummyPurchasableEntity>;
```
gmjosack, line 16: any reason these are capitalized? no other methods have capital case.

docs/script-api.md
```diff
@@ -5060,6 +5101,7 @@ For reference, the available `as_<typename>` functions are listed below:
 - as_dmspawning
 - as_door
 - as_drill
+- as_dummyPurchasableentity
```
gmjosack, line 5104: looks like docs need to be regenerated

src/game_api/entities_items.hpp
```diff
@@ -65,14 +75,24 @@ class VladsCape : public Cape
...
+class Mattock : public Purchasable
 {
   public:
     uint8_t remaining;
 };
 
-class Gun : public Movable
-{
+class Gun : public Purchasable
+{ // BIG NOTE: scepter is not a Purchasable, but that's the only Gun that isn't so i din't bother to make new type for it
```
gmjosack, line 95: Wouldn't this expose a non existent `buy` virtual for the scepter then? I imagine anyone calling that would be undefined and bad?

Mr-Auto: only stuff in src/game_api/script is exposed to lua, and also we decide what to expose, it's not automatic

until i know this function can be used in some way or another, there is no reason to add another pointless type for one entity, i added the rest only cuz it was relatively easy

src/game_api/level_api.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "game_allocator.hpp"
+#include "imgui.h"
```
Malacath-92, line 4: Why is this here?

src/game_api/level_api.hpp
```diff
@@ -458,9 +460,17 @@ struct LevelGenSystem
     std::uint32_t spawn_room_y;
     float spawn_x;
     float spawn_y;
-    DoorCoords* exit_doors_locations;
-    void* unknown37;
-    void* unknown38;
+    union
+    {
+        std::vector<ImVec2> exit_doors;
```
Malacath-92, line 465: Oh, I see it's for this. I'd prefer if we don't depend on imgui on the game types, that will make it more annoying later to split just the types and functionality (for pulling out types into their own headers as optional additions to the .dll interface).

Mr-Auto: I used this just cuz we already have this in API, exposed etc.

Malacath-92: Well I will have to remove it later either way.

Mr-Auto: ok, then i will make some quick XY struct

Malacath-92: Works for me  :)
And sorry, I didn't mean to be rude just there. It made perfect sense what you did without the context.

src/game_api/script/usertypes/entities_activefloors_lua.cpp
```diff
@@ -120,6 +118,8 @@ void register_usertypes(sol::state& lua)
         sol::base_classes,
         sol::bases<Entity, Movable>());
 
+    falling_platform["emitted_light"] = &FallingPlatform::timer; // wrong name, kept for backwards compatibility
```
Dregu, line 121: When a mistake is this obvious, I think it's the users fault for using it and not saying anything, so we can break backwards compatibility.

src/game_api/entity.cpp
```diff
@@ -716,3 +716,16 @@ void Entity::set_pre_collision2(std::uint32_t reserved_callback_id, std::functio
...
+{
+    if (items.size)
+        return std::span<uint32_t>(items.uids().begin(), items.uids().end());
+
+    return {};
+}
+
+void Movable::light_on_fire_broken()
+{
+    this->light_on_fire(0x64);
```
Malacath-92, line 730: Why exactly `0x64`, maybe put the answer in a comment?

src/game_api/level_api.hpp
```diff
@@ -499,6 +508,8 @@ struct LevelGenSystem
 
     uint32_t get_procedural_spawn_chance(uint32_t chance_id);
     bool set_procedural_spawn_chance(uint32_t chance_id, uint32_t inverse_chance);
+
+    ~LevelGenSystem(){}; // cuz it was complaining
```
Malacath-92, line 512: Where was it complaining? Are you invoking this function somewhere? Otherwise it would not complain, that needs to be fixed.

src/game_api/math.hpp
```diff
@@ -89,12 +89,114 @@ struct AABB
     float bottom{0};
 };
 
+struct XY
```
Malacath-92, line 92: This should probably be `Vec2` or something like that. No one in the history of linear algebra code called this `XY`  😜
If you want to make sure it doesn't clash with anything else put it into a `Spel2` namespave or something along those lines.

src/game_api/script/usertypes/entities_monsters_lua.cpp
```diff
@@ -269,6 +269,8 @@ void register_usertypes(sol::state& lua)
         &Shopkeeper::name,
         "shotgun_attack_delay",
         &Shopkeeper::shotgun_attack_delay,
+        "have_key",
```
Malacath-92, line 272: Should this not be `has_key`?

src/game_api/entity.cpp
Mr-Auto: The game normally lights you up for 100 frames, this is hard coded value.

I guess there could be some sources that set it to different times, but. This is a broken version of this function that shouldn't be used, kept for backwards compatibility. It doesn't show in the doc

src/game_api/level_api.hpp
Mr-Auto: When i added the `union` compiler started throwing `warning C4624: 'LevelGenSystem': destructor was implicitly defined as deleted`, i don't get why in this scenario it cares about the deconstructor, unless there is some code that calls it

src/game_api/entity.cpp
Malacath-92: Sorry for the confusion, I got why. My point was just that it's better to put a comment for future us when we're not currently looking at a PR. Magic numbers are not cool.

src/game_api/level_api.hpp
Malacath-92: It does sound like there is code that calls it, otherwise it should complain. If you look for the error in the Output window there should be more info.

Mr-Auto: There is nothing more the in the output, just the standard `the following warning is treated as an error` and then the warning that i posted on `level_api.hpp(511)`

Before that it builds the render_api which has only one line ralated to the level_api and it's `texture_id = State::get().ptr_local()->current_theme->get_dynamic_texture(texture_id);`

src/game_api/math.hpp
Mr-Auto: where does the '2' come from?

Malacath-92: It has two elements. As opposed to a `Vec3` which has 3 elements, etc. Same goes for matrices that you'd call `Mat2` and not `XXXYYXYY`

Mr-Auto: btw. i was thinking about constructor and split function from AABB, but not sure which corners i should choose :thinking:

Mr-Auto:
btw. i don't know how to fix the problem with generate.py :(

Malacath-92:
I'm gonna pull the PR and check for the two issues.

PR #193, opened 2022-02-14
Exports, console QoL and get_player

Dregu:
- Adds `import("script/id")` to get the `exports` table from other loaded scripts
- Adds `P`, `E` and `U` globals to console env to get player1 and selected entity, cause we're lazy
- Adds `get_player(1)` to get player (or ghost) by slot, cause `players[1]` is ambiguous


src/game_api/script/lua_vm.cpp
```diff
@@ -142,8 +142,48 @@ end
...
+                    if (player->inventory->player_slot == slot - 1)
+                        return sol::make_object_userdata(lua, player);
+                }
+            }
+            return sol::nil;
+        });
+    /// Returns PlayerGhost with this player slot 1..4
+    lua["get_playerghost"] = [&lua](int8_t slot) -> PlayerGhost*
+    {
+        for (auto uid : get_entities_by_type(to_id("ENT_TYPE_ITEM_PLAYERGHOST")))
```
Mr-Auto, line 179: Can you use get_entities_by with mask? it's faster then just the type

Also the game doesn't seam to care about the ghost, so there is no ghost that is assigned to slot, it's more of a: if ghost have this slot it will take inputs from that slot (i don't mean the code is wrong or anything, just wanted to clarify)

src/game_api/script/lua_vm.cpp
```diff
@@ -142,8 +142,48 @@ end
...
+            for (auto player : get_players())
+            {
+                if (player->inventory_ptr->player_slot == slot - 1)
+                    return sol::make_object_userdata(lua, player);
+            }
+            if (or_ghost)
+            {
+                for (auto uid : get_entities_by_type(to_id("ENT_TYPE_ITEM_PLAYERGHOST")))
+                {
+                    auto player = lua["get_entity"](uid).get<PlayerGhost*>();
```
Mr-Auto, line 169: Also, is there a reason for `lua["get_entity"]` instead of `get_entity_ptr` and `as<PlayerGhost>()` ?

src/game_api/script/lua_vm.cpp
Dregu: Oh right you did say usings masks anyway is faster

src/game_api/script/lua_vm.cpp
Dregu: I guess there isn't. First I thought this would cast it nicely for me but turns out it's not even easier

PR #195, opened 2022-02-17
Auto docs

Malacath-92:
Automatically generate API docs and push to Wiki on any push to main


src/game_api/script/lua_vm.cpp
```diff
@@ -254,6 +254,7 @@ end
         return backend->cbcount++;
     };
     /// Clear previously added callback `id` or call without arguments inside any callback to clear that callback after it returns.
+    // lua["clear_callback"] = [](sol::optional<CallbackId> id) -> void {};
```
Mr-Auto, line 257: Would probably be better to just make two separate functions for this and save overload to variable as we always do

docs/script-api.md
```diff
@@ -1,6125 +1,2 @@
...
-- [`float fade_away_counter`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=fade_away_counter) &Leaf::fade_away_counter
-\
-counts to 100.0 then the leaf fades away
-- [`int swing_direction`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=swing_direction) &Leaf::swing_direction
-- [`bool fade_away_trigger`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=fade_away_trigger) &Leaf::fade_away_trigger
-### `AcidBubble`
-Derived from [`Entity`](#entity) [`Movable`](#movable)
-- [`float speed_x`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=speed_x) &AcidBubble::speed_x
-- [`float speed_y`](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=speed_y) &AcidBubble::speed_y
-- [`float float_counter`](https://github.com/spelunky-fyi/overl
```
Mr-Auto: I thought the idea was to keep this and add wiki?

Malacath-92: There's no need to keep this though. Any good reason?

Malacath-92: Wait... Really?

Malacath-92: Well that sucks, because the Wiki is better at everything else... If we structure it well you won't have any difficulties to find the right path though on your own.

Dregu: The hamburger menu was the only way to find anything in script-api, taking it away completely is not a good option before a new actual doc platform like Slate is up with proper search across all pages.

Mr-Auto: Most of my interaction with script-api is checking the function parameters with simple Ctrl+F. 
Even with best tree structure the wiki will be a annoying. But that could be just 'me' problem

Dregu: Wiki will be a better alternative for browsing. Marginally, global functions is still 47 pages printed compared to 207 pages of script-api.

PR #197, opened 2022-02-23
Improve build time, reduce connections between files etc.

Mr-Auto:


Dregu:
The goal in the title is a nice thought, but CI is saying build on this took a bit longer than my pr...

Mr-Auto:
@Dregu i didn't actually test the full rebuild time, the main goal was to reduce the connections, so when you change one file you don't need to rebuild 65 other

src/injected/ui.cpp
```diff
@@ -383,7 +383,7 @@ void hook_savegame()
     {
         register_on_write_to_file([](const char* backup_file, const char* file, void* data, size_t data_size, WriteToFileOrig* original)
                                   {
-                                      if (file == "savegame.sav"sv and options["disable_savegame"])
+                                      if (strcmp(file, "savegame.sav") == 0 and options["disable_savegame"])
```
Malacath-92, line 386: why tho?

Mr-Auto: `file` is `const char*`, so before it was comparing pointers not strings

Malacath-92: What is was doing before was comparing a string_view to a const char*, which is perfectly well defined behavior

Mr-Auto: oh, is that way better than `strcmp`? 
cuz this is the only use of `string_view` in the entire UI

Malacath-92: It's not better or worse, just more C++ as opposed to using the C stl
I mainly thought the diff was unnecessary.

PR #199, opened 2022-03-02
Add basic settings API

Malacath-92:


Dregu:
Oh right, `DAMSEL` is not used anywhere in the api or ui (except the sound guy didn't get the message), lets not call the setting damsel either.

Malacath-92:
These enum values are generated from the binary, they are called like that internally. Do we really want to fiddle with that?

Malacath-92:
Also what's the reference with "except the sound guy didn't get the message" exactly?

Dregu:
In entities and textures they are called pets, but in sounds they are damsels.. There's even one for petting a damsel, which is pretty funny. I didn't know there were any other internal references to damsel.

PR #200, opened 2022-03-05
more drawing functions from imgui

Dregu:
Is this api ok?

### DIFF
--- a/src/game_api/script/usertypes/gui_lua.cpp
+++ b/src/game_api/script/usertypes/gui_lua.cpp
@@ -88,6 +88,44 @@ void GuiDrawContext::draw_rect_filled(AABB rect, float rounding, uColor color)
 {
     draw_rect_filled(rect.left, rect.top, rect.right, rect.bottom, rounding, color);
 }
+void GuiDrawContext::draw_poly(std::vector<Vec2> points, float thickness, uColor color)
+{
+    backend->draw_list->PathClear();
+    for (auto point : points)
+    {
+        ImVec2 a = screenify({point.x, point.y});
+        backend->draw_list->PathLineToMergeDuplicate(a);
+    }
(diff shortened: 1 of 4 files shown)


src/game_api/script/usertypes/gui_lua.cpp
```diff
@@ -88,6 +88,44 @@ void GuiDrawContext::draw_rect_filled(AABB rect, float rounding, uColor color)
...
+void GuiDrawContext::draw_bezier(std::vector<Vec2> points, float thickness, uColor color)
+{
+    if (points.size() == 3)
+    {
+        ImVec2 a = screenify({points[0].x, points[0].y});
+        ImVec2 b = screenify({points[1].x, points[1].y});
+        ImVec2 c = screenify({points[2].x, points[2].y});
+        backend->draw_list->AddBezierQuadratic(a, b, c, color, thickness);
+    }
+    else if (points.size() == 4)
```
Mr-Auto, line 120: Wouldn't be better to make 2 functions (3/4 parameters instead of vector) and overload instead of relaying on the number of elements in vector?

Also, you can probably cast Vec2 to ImVec2 as it's structurally the same (and the screenify only accesses x and y) or make overload for screenify to use Vec2

Dregu: And why would it be better?

Mr-Auto: Just more clear what the function needs, it's also more fail proof if someone wants to use the BezierCubic but makes mistake in the code and give it 3 element table or nil's one of the 4 elements by accident

I didn't do much for drawing in the API but it there advantage of keeping the Vec2 in a table in lua?
cuz if it's end up being: make table just to pass the parameters to the function, there is a little to gain from having a vector for the coordinates

PR #201, opened 2022-03-06
Growable entities

Mr-Auto:
I did the game function only for the `grow_chain_and_blocks` as it's a little complicated, with all the spawn_over and prng.

Each of growables have similar function, but it's simple enough to make from scratch and they don't fix the animation_frame


src/game_api/layer.hpp
```diff
@@ -4,6 +4,9 @@
 
 class Entity;
 
+constexpr uint32_t g_level_max_x = 0x56;
```
Malacath-92, line 7: Should be `inline`

Mr-Auto: got it

PR #202, opened 2022-03-10
Fix `move_grid_entity`

Mr-Auto:
Mostly fix if entities attached to the moved entities are also floors

Also this function now does the same thing as `move_entity` (except moving liquids), but doesn't break floor collision


src/game_api/layer.cpp
```diff
@@ -93,8 +93,8 @@ Entity* Layer::spawn_entity_over(ENT_TYPE id, Entity* overlay, float x, float y)
 
 Entity* Layer::get_grid_entity_at(float x, float y)
 {
-    const uint32_t ix = static_cast<uint32_t>(x + 0.5f);
```
Malacath-92, line 96: Why this change? I would need to find the offset to convince you perhaps, but this was based on exactly the disasm. I believe the behavior is essentially the same, but maybe not, so curious what your reasoning was.

src/game_api/layer.cpp
```diff
@@ -173,3 +173,32 @@ Entity* Layer::spawn_apep(float x, float y, bool right)
...
+void Layer::move_grid_entity(Entity* ent, float x, float y, Layer* dest_layer)
+{
+    move_grid_entity(ent, static_cast<uint32_t>(std::round(x)), static_cast<uint32_t>(std::round(y)), dest_layer);
+}
+
+void Layer::move_grid_entity(Entity* ent, uint32_t x, uint32_t y, Layer* dest_layer)
+{
+    if (ent)
+    {
+        if (ent->type->search_flags & 0x100) // MASK.FLOOR
```
Malacath-92, line 186: What's up here? Why move only floors?

Mr-Auto: Only floors are grid entities as far i know

Malacath-92: But users can spawn anything as a grid entity via the API, right?

src/game_api/layer.cpp
Mr-Auto: Just to make it clear what it does instead of adding some magic number. Nothing important, i can change it back if you insist 

Also apparently the standard covers more situations, the only difference i found is rounding -0.0 (the +0.5f rounds to 0.0 instead of -0.0), not really important in this situation of course

src/game_api/layer.cpp
Mr-Auto: Nope, only get_grid_entity and move_grid_entity access `layer->grid_entities`, and it's not exposed to lua, spawn_grid_entity uses the standard spawn function

Malacath-92: But doesn't that standard spawn function, when called with the right params, place the spawned entity into the grid? Not in our code but in game code.

Mr-Auto: Don't think so? what would be the reason for that? spawning the floor with standard parameters adds it to the grid
also have you tested if that even does anything when adding non floor entity to the grid?

Malacath-92: No, that's why I'm asking you. You're currently working with this so it's only natural that you would answer questions about it  😉

PR #203, opened 2022-03-12
UI stuff

Mr-Auto:


Mr-Auto:
I understand if this gets deleted, at the end i feel like I've done absolutely nothing with this :/

src/game_api/script/lua_vm.cpp
```diff
@@ -190,7 +190,8 @@ end
         return nullptr;
     };
     /// Provides a read-only access to the save data, updated as soon as something changes (i.e. before it's written to savegame.sav.)
-    lua["savegame"] = savedata();
+    lua["savegame"] = []() -> SaveData*
```
Malacath-92, line 193: This was a value before, now it's a function. Seems like a breaking change, or am I missing something?

Mr-Auto: Sorry, didn't realized it's call the funciton not reference the function

Malacath-92 (approved):

PR #207, opened 2022-04-02
Fix `grow_vines` function

Mr-Auto:


src/game_api/level_api.cpp
```diff
@@ -1923,27 +1923,30 @@ void grow_vines(LAYER l, uint32_t max_lengh, AABB area, bool destroy_broken)
 
     const uint32_t start_x = static_cast<uint32_t>(area.left + 0.5f);
     uint32_t end_x = static_cast<uint32_t>(area.right + 0.5f);
-    const uint32_t start_y = static_cast<uint32_t>(area.bottom + 0.5f);
-    uint32_t end_y = static_cast<uint32_t>(area.top + 0.5f);
+    uint32_t start_y = static_cast<uint32_t>(area.top + 0.5f);
+    int end_y = static_cast<uint32_t>(area.bottom + 0.5f);
```
Malacath-92, line 1927: Cast to uint32 then implicit cast to int? Maybe just use int32 here?

Malacath-92 (changes requested):

src/game_api/level_api.cpp
Mr-Auto: it should be `uint32_t`, fixed, thanks

PR #211, opened 2022-04-21
Small render callbacks

Malacath-92:
- Added a `RENDER_POST_DRAW_DEPTH`, same concept as `RENDER_PRE_DRAW_DETPH`
- Added `set_pre_render` for entities, which allows blocking the original render
- Added `set_post_render` for entities
- Added `WORLD_SHADER` enum
- Added `shader` param to `VanillaDrawContext::draw_world_texture`
- Added `online_safe` option for script meta
- Linked entity hooks to uid for making them work in online MP


Malacath-92:
> Have you tested the shades in the draw screen texture as well?

No I have not, all these shaders are not compatible with that though because the screen draw functions use screen coordinates and all those shaders expect world coordinates.

PR #213, opened 2022-04-22
Sparktrap hack to have speed and direction per spark

Mr-Auto:


src/game_api/entities_items.hpp
```diff
@@ -289,6 +289,12 @@ class Spark : public Flame
     float next_size;
     /// very short timer before next size change, giving a pulsing effect
     uint8_t size_change_timer;
+    uint8_t extra_padding[3];
+
+    /// This is cusome variable, you need to use `activate_sparktraps_hack` to use it
+    float speed;
+    /// This is cusome variable, you need to use `activate_sparktraps_hack` to use it
+    float distance;
```
Malacath-92, line 297: This is a bit of a yolo, but I guess considering how the game allocates entities it's fine  😆 
Have you just double-checked that the size of a sparktrap is not right at the border of the bucket size?

src/game_api/rpc.cpp
```diff
@@ -806,7 +806,11 @@ void modify_sparktraps(float angle_increment, float distance)
     static float* new_array;
     if (new_array == nullptr)
     {
-        const auto offset = get_address("sparktrap_angle_increment");
+        const auto offset = get_address("sparktrap_angle_increment") + 4;
```
Malacath-92, line 809: Make this static?

src/game_api/rpc.cpp
```diff
@@ -806,7 +806,11 @@ void modify_sparktraps(float angle_increment, float distance)
...
     if (new_array == nullptr)
     {
-        const auto offset = get_address("sparktrap_angle_increment");
+        const auto offset = get_address("sparktrap_angle_increment") + 4;
+
+        if (read_u8(offset - 1) != 0x89) // check if sparktraps_hack is active
+            return;
+
         const int32_t distance_offset = 0xF1;
         new_array = (float*)alloc_mem_rel32(offset + 4, sizeof(float) * 2);
```
Malacath-92, line 815: Also this offset is used here with `+4` again. What's the deal?

Mr-Auto: the offset for jump/access memory is distance from past the expression, `alloc_mem_rel32` looks for the available memory above the given offset, so i have to correct for the 4 bits of the actual offset of the expression (otherwise if it would get free memory at exactly `offset - 32bit` it would be more then signed 32bit value cousin overflow)

hope that this explanation makes sense :stuck_out_tongue_closed_eyes:

src/game_api/rpc.cpp
Mr-Auto: aren't addresses in get_address cached?

src/game_api/entities_items.hpp
Mr-Auto: Yeah, we have quite a bit of room in fact

src/game_api/rpc.cpp
Malacath-92: I guess, but they still need to be looked up in a map. Idk what's more efficient, static vars or map lookups.

src/game_api/rpc.cpp
Malacath-92: Doesn't really clear up why the other code that uses offset does not need to be modified for me. Was this code previously wrong?

src/game_api/rpc.cpp
Mr-Auto: Static should be faster, not that it matters in such non-frequent used function

src/game_api/rpc.cpp
Malacath-92: Ah yes, the missing piece for me was that I didn't notice that it was a `find_after_inst`, now it all makes sense.

PR #221, opened 2022-05-13
Expose render info

Malacath-92:
Just need to regen the emmylua


src/game_api/math.hpp
```diff
@@ -283,6 +283,26 @@ struct Quad
         return *this;
     }
 
+    Quad& flip_horizontal()
```
Mr-Auto, line 286: Shouldn't this be `flip_horizontally` to match with the other function?
or is this some English exception that i don't know of?

Mr-Auto (changes requested):

PR #223, opened 2022-05-14
Expose powerups in inventory

Mr-Auto:


src/game_api/items.hpp
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "entity.hpp"
+#include "entities_chars.hpp"
```
Malacath-92, line 3: Wouldn't it be better to move Inventory to its own header? Moving from entity.hpp to entities_chars.hpp ends up introducing heavier dependencies, right?

Mr-Auto: kind of, but also i don't understand why it needs to be in `items.hpp` in first place, it throws some random errors when it's not

Issue #219, opened 2022-05-10
Add a callback when grid entities are updated


jaythebusinessgoose:
After a grid entity is destroyed (eg, during an explosion), the grid entities within 2 tiles are updated. This update includes updating the decorations on floor tiles.

It would be nice to have hooks into whatever method(s) called here, so that things can be fixed after the game makes these updates.

PR #229, opened 2022-06-01
Floor update hooks and adventure seed

Dregu:
`set_pre_update` and `set_post_update`, called before and after a floor neighbor is destroyed and I guess on levelgen too. For #219.

Here's a random example
```lua
  set_post_update(ent.uid, function(me)
    me:set_texture(TEXTURE.DATA_TEXTURES_FLOOR_JUNGLE_0)
    for i,v in ipairs(entity_get_items_by(me.uid, ENT_TYPE.DECORATION_GENERIC, MASK.DECORATION)) do
      local deco = get_entity(v)
      deco:set_texture(TEXTURE.DATA_TEXTURES_FLOOR_VOLCANO_0)
    end
  end)
end, SPAWN_TYPE.LEVEL_GEN, MASK.FLOOR, ENT_TYPE.FLOOR_GENERIC)
```

Or a tilecode to use multiple generic floors in one theme

```lua
define_tile_code("floor_generic_tidepool")
set_pre_tile_code_callback(function(x, y, layer)
    local uid = spawn_grid_entity(ENT_TYPE.FLOOR_GENERIC, x, y, layer)
    set_post_update(uid, function(me)
        me:set_texture(TEXTURE.DATA_TEXTURES_FLOOR_TIDEPOOL_0)
        for i,v in ipairs(entity_get_items_by(me.uid, ENT_TYPE.DECORATION_GENERIC, MASK.DECORATION)) do
            local deco = get_entity(v)
            deco:set_texture(TEXTURE.DATA_TEXTURES_FLOOR_TIDEPOOL_0)
        end
    end)
    return true
end, "floor_generic_tidepool")
```

Catch a screen change before it happens and divert to somewhere else. This skips transitions completely:

```lua
set_callback(function()
    if state.screen_next == SCREEN.TRANSITION then
        state.screen_next = SCREEN.LEVEL
    end
end, ON.PRE_LOAD_SCREEN)
```

The `load_screen` function is what the game calls when setting `state.loading == 2` to apply screen_next etc, but you can call it in any state. It's kinda janky and crashy though, probably shouldn't keep it. The callbacks around it are more useful.

Also fixed some funny conditions in render hooks.


src/game_api/rpc.cpp
```diff
@@ -1758,3 +1758,33 @@ void change_poison_timer(int16_t frames)
...
     }
 }
+
+void set_adventure_seed(int64_t first, int64_t second)
+{
+    static size_t offset = 0;
+    if (offset == 0)
+    {
+        auto memory = Memory::get();
+        offset = memory.at_exe(0x22dd9d40); // TODO
```
Dregu, line 1768: I don't know

Dregu: Oh that was easy

Malacath-92 (changes requested):

docs/game_data/spel2.lua
```diff
@@ -890,6 +890,24 @@ function set_on_player_instagib(uid, fun) end
 ---@return CallbackId?
 function set_on_damage(uid, fun) end
 ---Returns unique id for the callback to be used in [clear_entity_callback](#clear_entity_callback) or `nil` if uid is not valid.
+---Sets a callback that is called right before a floor is updated (by killed neighbor), return `true` to skip the game's neighbor update handling.
```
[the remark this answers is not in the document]
Dregu: Yeah maybe I should call it floor update. This does have the option to hook some other virtual on movables, but I don't know what virtual that would be.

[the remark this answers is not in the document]
Mr-Auto: For movable that would be statemachine, that we already have a hook for, if i understand correctly

src/game_api/search.cpp
```diff
@@ -1600,6 +1600,29 @@ std::unordered_map<std::string_view, AddressRule> g_address_rules{
...
+            .offset(0x8)
+            .decode_call()
+            .at_exe(),
+    },
+    {
+        "load_screen_func"sv,
+        PatternCommandBuffer{}
+            .find_inst("\x8b\x49\x0c\x41\x8b\x47\x10"sv)
+            .at_exe()
+            .function_start(),
```
Mr-Auto, line 1616: Why to patterns for the same function?

Mr-Auto (changes requested):

src/game_api/search.cpp
Dregu: I guess I didn't know who I liked more, the call or the body

PR #231, opened 2022-06-02
Add "Filtered seeds" script to auto-reset for good 1-2 shops

amclark:
Select which items you want and any time you reset it will put you into a new random seed that matches your requirements. Also supports requiring a Kali altar on 1-2 and optionally searching in back layer shops for items as well (e.g. for plasma cannon). Thinking this will be useful for practice or casual runs.


amclark:
per convo with Dregu in #s2-modding-help a seedfinder could be useful but example scripts isn't the right spot for it

PR #232, opened 2022-06-03
Add `update_liquid_collision_at`

Mr-Auto:
Not sure if this should be two separate functions or just one as i did right now


Dregu:
Yep works great for with destroying floors, now with 1000x less rubble

PR #233, opened 2022-06-06
Viewports

Dregu:
- Added magic viewports support and enabled it by default
- Added new main menu UI (instead of the tabbed floating window you have a tiny menubar) and enabled it by default
- You can still drag individual windows out of the menu if you want
- Removed old stacked interfaces since the new menu UI is objectively better
- I probably did some dumb code again, but since this is just UI I think I'll release it for testing and fix later


Dregu:
Apparently I broke playlunky console somehow, have to fix that later.

PR #234, opened 2022-06-06
Custom behavior

Malacath-92:
Gonna write some stuff about it later


Malacath-92:
So basically this lets you write a custom state machine for an entity. The rock example is obviously overkill for state machines, it's just a proof of concept. There's a bit of confusion in naming now because someone decided to call parts of this behavior while it was initially called state  😛
`MovableBehavior` has all the customization points, they are explained on their relevant setters in `behavior_lua.cpp`

I also added access to the `MemHeap::Alloc` and `MemHeap::Free` (I think iojon used these names back in the day) functions through `custom_allocator.cpp` and wrapped stl containers to use this. Then I split all the containers up so we introduce less dependencies when we fiddle with that stuff.

I added a tiny utility to make debugging in VS easier, can no put `to_name(0x1)` into the watch window to get the actual name as a string.

I have no idea what happened with movable.hpp

Nothing else to say I think

docs/src/includes/_globals.md
```diff
@@ -1707,6 +1707,159 @@ Show a message that looks like a level feeling.
...
 
+## Movable Behavior functions
+
+
+### add_behavior
+
+
+> Search script examples for [add_behavior](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=add_behavior)
+
+#### nil add_behavior([Movable](#Movable) movable, MovableBehavior behavior)
```
Mr-Auto, line 1718: I don't think we ever have entity as parameter, tend to use uid or at least function for specific type/class

Mr-Auto (changes requested):

docs/src/includes/_globals.md
Malacath-92: It just makes more sense tho when we actually have restrictions on what types can be passed. Instead of the stuff that I wrote before like `set_post_statemachine` that just break if you pass a non-movable (or maybe doesn't right now, but with this you'll get an error at least instead of a silent failure)

Mr-Auto: But if it's only for movable, why not make it a function for that class?

Malacath-92: Because we already have an issue with insane dependencies and huge compile times. I'd rather just keep the functions separate. If we prefer to we can still make those functions bind directly to the Movable type on the Lua side, but keep the different "logical modules" separate on the C++ side. Should I take a look at doing that?

Malacath-92: > also there is internal function is_movable() so if you would go with uid, you could use it check before doing any changes for that entity
Yes, that's what we use in other cases, and afaik that results in a silent failure right now, which we shouldn't be satisfied with either.

Mr-Auto: > Because we already have an issue with insane dependencies and huge compile times. I'd rather just keep the functions separate.

To have `Movable` as parameter you need the reference of Movable anyway, so i don't quite get what you mean by this.

>  If we prefer to we can still make those functions bind directly to the Movable type on the Lua side, but keep the different "logical modules" separate on the C++ side. Should I take a look at doing that?

I think yes? if i understand it correctly, i remember you mentioning moving the API function away from the game structs. Would it be possible to make a new class that inherit from in this example Movable, just to have all the functions that we made but for the lua it would still bind it to Movable?

Malacath-92: If I move this stuff to `Movable.hpp` then every time I wanna change anything about it all dependents on that file have to recompile. And that file has maaaaany dependents, so compile times will go up. So keeping it separate will have less adverse effects on compile times.

I think we can bind free functions to types in Lua as well so long as the first parameter matches. I'll take a look.

Malacath-92:
@Mr-Auto See my two latest commits, the functions are now bound to `Movable`, `MovableBehavior` and `CustomMovableBehavior` respectively. I hope you see now why I think how little difference this makes, but I don't mind either way.
I will fix docs next.

PR #235, opened 2022-06-07
Better way for refreshing callbacks? potentially sync them for online?

Mr-Auto:
I left the update function for testing


Dregu:
Heh I added a hook for the same function a few days ago, but didn't want to mix it up with the ui stuff yet. Didn't make anything to support better updates for the old callbacks tho.

Dregu:
Getting the updates synced is probably a big thing for online, but also using more of the thread-local versions of state/layer. When rollback or new level happens, things are pointing to an old State. Or something.

Mr-Auto:
> Getting the updates synced is probably a big thing for online, but also using more of the thread-local versions of state/layer. When rollback or new level happens, things are pointing to an old State. Or something.

Yeah, i want to look into it as well, just, not have much time lately

src/game_api/rpc.cpp
```diff
@@ -1846,3 +1847,25 @@ void game_log(std::string message)
     const static auto log_stream = (std::ofstream*)read_i64(get_address("game_log_stream"));
     game_log_fun(log_stream, message.c_str(), nullptr, LogLevel::Info);
 }
+
+using OnStateUpdate = void(StateMemory*);
+OnStateUpdate* g_state_update_trampoline{nullptr};
+void StateUpdate(StateMemory* s)
+{
+    g_state_update_trampoline(s);
```
Dregu, line 1855: Could you add a pre-update callback too (PRE_EVERYFRAME?) that can cancel the trampoline by returning true (like many other callbacks), or I can do it after this is merged. I'm not sure why yet, but I think it might be useful.

src/game_api/script/lua_backend.hpp
```diff
@@ -94,6 +94,7 @@ enum class ON
     PRE_LOAD_SCREEN,
     POST_LOAD_SCREEN,
     DEATH_MESSAGE,
+    EVERYFRAME,
```
Dregu, line 97: I'd still rather call this event UPDATE (and then PRE_UPDATE for the other thing I asked)

Mr-Auto: I still think it should have "frame" in the name, to signify that it's executed with the game framerate

Dregu: Well I think "every frame" is just confusing if you're playing at 300fps. A comment should make it clear enough that it runs at 60fps, after every state update

Mr-Auto: But you can't play at 300fps, unless i don't know about some special tech?

you can play on monitor with some high refresh rate, but that's not the same

Dregu: My game runs at 300fps and I see 144 of those frames on my monitor. Calling something 60fps "every frame" is neither of those things. If you really have to have FRAME in the name, call it LOGICFRAME or something.

Mr-Auto: With that logic ON.FRAME is very bad name and GAMEFRAME even worst

Dregu: Ok lets make everything bad cause other things are bad

Dregu: I'm not trying to argue that this event is not a "frame", just like those other things are not "frames", I'm just saying this ain't fucking EVERY FRAME

Dregu:
I other news, I think my game stopped immediately crashing when playing this PR online. I don't know what happened. The rando I made specifically for online testing is not totally synced though, but that might be because the script is bad.

src/game_api/rpc.cpp
Mr-Auto: Done

src/game_api/rpc.cpp
```diff
@@ -40,19 +40,19 @@
 #include "thread_utils.hpp"     // for OnHeapPointer
 #include "virtual_table.hpp"    // for get_virtual_function_address, VIRT_FUNC
 
-uint32_t setflag(uint32_t flags, int bit) // shouldn't we change those to #define ?
+inline uint32_t setflag(uint32_t flags, int bit) // shouldn't we change those to #define ?
```
Malacath-92, line 43: Please don't put inline on functions unless you actually need inline linkage.

src/game_api/script/lua_backend.hpp
```diff
@@ -94,6 +94,8 @@ enum class ON
     PRE_LOAD_SCREEN,
     POST_LOAD_SCREEN,
     DEATH_MESSAGE,
+    PRE_UPDATE,
+    UPDATE,
```
Malacath-92, line 98: Shouldn't this be `POST_UPDATE`?

src/game_api/script/lua_backend.hpp
```diff
@@ -288,6 +290,7 @@ class LuaBackend
 
     void clear();
     void clear_all_callbacks();
+    bool frame_update(StateMemory* state_mem);
```
Malacath-92, line 293: Do we really gain anything from this rename?

src/game_api/script/lua_vm.cpp
```diff
@@ -412,7 +412,6 @@ end
             if (!import_backend->get_enabled())
             {
                 import_backend->set_enabled(true);
-                import_backend->update();
```
Malacath-92, line 415: Why was this previously updating? Presumably it did that to run any code at global scope. So is it really safe to stop doing this?

src/game_api/script/lua_vm.cpp
```diff
@@ -1902,6 +1902,11 @@ end
     // DEATH_MESSAGE
     // Params: `STRINGID id`
     // Runs once after death when the death message journal page is shown. The parameter is the STRINGID of the title, like 1221 for BLOWN UP.
+    // PRE_UPDATE
+    // Runs before the State is updated, runs always (menu, settings, camp, game, arena, online etc.) with the game engine, thypically 60FPS
```
Malacath-92, line 1906: Type: typhically -> typically

src/game_api/script/lua_vm.cpp
```diff
@@ -1902,6 +1902,11 @@ end
...
+    // Return behavior: return true to stop futher PRE_UPDATE callbacks from executing and don't update the state (this will essentially freeze the game engine)
+    // UPDATE
+    // Runs right after the State is updated, runs always (menu, settings, camp, game, arena, online etc.) with the game engine, thypically 60FPS
```
Malacath-92, line 1909: Ditto

src/game_api/state.cpp
```diff
@@ -238,22 +239,36 @@ State& State::get()
...
+    if (g_state_ptr)
+        return g_state_ptr;
+
     OnHeapPointer<StateMemory> p(read_u64(location));
     return p.decode();
 }
 
 StateMemory* State::ptr_local() const
 {
+    if (g_state_ptr)
```
Malacath-92, line 260: Why? This function is getting another state pointer by design, so you can't take the one that you got from `State::ptr()`

src/game_api/state.hpp
```diff
@@ -51,16 +51,19 @@ struct StateMemory
     float fadevalue; // 0.0 = all visible; 1.0 = all black
     uint32_t fadeout;
     uint32_t fadein;
-    uint32_t loading_black_screen_timer; // if state.loading is 1, this timer counts down to 0 while the screen is black (used after Ouroboros, in
-                                         // credits, ...)
+    uint32_t loading_black_screen_timer; // if state.loading is 1, this timer counts down to 0 while the screen is black (used after Ouroboros, in credits etc.)
+
+    /// True when you have control over your character
     uint8_t ingame;
+    /// True whenever you in an active game (basically everything except the main menu, character select etc.)
```
Malacath-92, line 58: Type: "whenever you in" -> "whenever you are in"

Malacath-92 (changes requested):

src/game_api/rpc.cpp
Mr-Auto: What's wrong with inline?

src/game_api/script/lua_backend.hpp
Mr-Auto: Dreagu requested change from everyframe to update, but makes sens to make it post_update

src/game_api/script/lua_backend.hpp
Mr-Auto: Not really, originally i left the update so i had to change this to something else

Also didn't want the confusion with the old way where the updated had to be called non-stop where now it's all based on hook

src/game_api/rpc.cpp
Malacath-92: `inline` declares a function as having `inline` linkage, it does not tell the compiler to inline the function. And you don't need `inline` linkage here from what I can see so it's completely unnecessary.

src/game_api/script/lua_vm.cpp
Mr-Auto: "run any code at global scope" - i don't quite get that

This will run the update out of sync, but i revert that for now

docs/game_data/lua_enums.txt
```diff
@@ -1763,11 +1764,13 @@ ON = {
   ONLINE_LOBBY = 29,
   OPTIONS = 5,
   POST_LEVEL_GENERATION = 112,
+  POST_LOAD_SCREEN = 127,
   POST_ROOM_GENERATION = 111,
   PRE_GET_RANDOM_ROOM = 113,
   PRE_HANDLE_ROOM_TILES = 114,
   PRE_LEVEL_GENERATION = 110,
   PRE_LOAD_LEVEL_FILES = 109,
+  PRE_LOAD_SCREEN = 126,
```
Mr-Auto, line 1773: Not sure why it didn't add the PRE/POST UPDATE here

Mr-Auto:
I don't know what to do with the update being called from multiple threads
@Malacath-92 not sure if you want to merge this to your branch maybe? i think I'm done with this PR

Dregu:
This has been in limbo for months but would be very beneficial for single player so... Does it actually break anything that currently barely works in online? Do we actually even care about online?

Dregu:
And clearly `state` etc should always point to the thread-local state in case the update gets called from another thread, so the changes made to it are also only in the thread-local state and the game handles the rest if that state ends up reality? So just set `lua["state"] = ptr_local` at the beginning of an update and back to ptr_main at the end or something?

Dregu:
Randonline actually runs in sync most of the time with some of these changes. The issues might be really tiny, like some thread advancing the main prng when it should be using local prng. Maybe we'll get this working this year.....

PR #237, opened 2022-06-12
UI tweaks and set_gravity

Dregu:


src/game_api/entity.cpp
```diff
@@ -838,3 +838,14 @@ uint32_t Movable::get_behavior()
     }
     return 0; // there is no id 0, but i can be wrong
 }
+
+void Movable::set_gravity(float gravity)
```
Dregu, line 842: Is this a stupid way to do this? Does everything blow up if you do this every frame for example?

src/game_api/entity.cpp
```diff
@@ -838,3 +838,14 @@ uint32_t Movable::get_behavior()
...
     return 0; // there is no id 0, but i can be wrong
 }
+
+void Movable::set_gravity(float gravity)
+{
+    hook_vtable<void(Movable*, float)>(
+        this,
+        [gravity](Movable* ent, [[maybe_unused]] float _gravity, void (*original)(Movable*, float))
+        {
+            original(ent, gravity);
```
Dregu, line 848: Is this a stupid way to do this? Does everything blow up if you do this every frame for example?

src/game_api/entity.cpp
```diff
@@ -838,3 +838,14 @@ uint32_t Movable::get_behavior()
...
+        },
+        0x53);
```
Dregu, line 867: Is this a stupid way to do this? Does everything blow up if you do this every frame for example?

PR #239, opened 2022-06-14
User data

Dregu:
I added some kind of a thing to save userdata for Players and the entities transferred with them just before unloading a level, and restore on post level gen. Might even work with funky HH chains.


docs/src/includes/_types.md
```diff
@@ -3106,6 +3106,7 @@ nil | [apply_metadata(int metadata)](https://github.com/spelunky-fyi/overlunky/s
 nil | [set_invisible(bool value)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_invisible) | 
 array&lt;int&gt; | [get_items()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_items) | 
 bool | [is_in_liquid()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=is_in_liquid) | Returns true if entity is in water/lava
+nil | [user_data()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=user_data) | Table containing arbitrary data, `nil` by default and will be cleared when the entity is destroyed
```
Malacath-92, line 3109: Something is wrong here, this is probably gonna be confusing. I haven't checked yet why this generates as it does tbh. Maybe we should add one of those comment-annotations in code?

src/game_api/script/lua_backend.cpp
```diff
@@ -177,6 +178,51 @@ CustomMovableBehavior* LuaBackend::make_custom_movable_behavior(std::string_view
...
+    return sol::nil;
+}
+void LuaBackend::set_user_data(Entity& entity, sol::object user_data)
+{
+    if (!user_datas.contains(entity.uid))
+    {
+        entity.set_on_dtor(
+            [this, &entity](Entity*)
+            {
+                /*auto state = State::get().ptr_local();
```
Malacath-92, line 204: Leftover testing code or still needed to finish up?

src/game_api/script/lua_backend.cpp
```diff
@@ -177,6 +178,51 @@ CustomMovableBehavior* LuaBackend::make_custom_movable_behavior(std::string_view
...
+    {
+        return user_datas[uid];
+    }
+    return sol::nil;
+}
+void LuaBackend::set_user_data(Entity& entity, sol::object user_data)
+{
+    if (!user_datas.contains(entity.uid))
+    {
+        entity.set_on_dtor(
```
Malacath-92, line 201: I think we need to store the returned hook id and clear the hook in `clear_all_callbacks` to make sure a reload is not gonna call stale code?

Malacath-92:
Oh wait, is this my branch  😅 
I can do the changes I commented then myself ig

src/game_api/script/lua_backend.cpp
Dregu: I tried doing it in the dtor first, which technically works but is stupid if the player doesn't have userdata set, only the carried item. You'd need to figure out in reverse if the item is being carried by a player.

Dregu: Unless of course you hooked the player dtor always, not just when it has userdata

src/game_api/script/lua_backend.cpp
Dregu: Probably, I didn't address the fucked up cleanup yet

src/game_api/script/lua_backend.cpp
Dregu: Actually things you are carrying might still get destroyed before the player, the other way is much safer.

docs/src/includes/_types.md
Dregu: That's not updated, the new doc seems fine with no type and no (). I changed it to store any type btw.

Dregu:
So I made an example how to combine these features, but it's currently bugged and the level/start callback fires after the skipped death screen, when it shouldn't. However whatever @Mr-Auto is doing might fix this, and we should look into using better (POST_LOAD_SCREEN?) logic for the callbacks anyway.

```lua
set_callback(function()
  players[1].user_data = {lives = 2}
  print(F"You have {players[1].user_data.lives} extra lives!")
end, ON.START)

set_callback(function()
  if state.screen_next == SCREEN.DEATH and players[1] and players[1].user_data.lives and players[1].user_data.lives > 0 then
    P.flags = clr_flag(P.flags, ENT_FLAG.DEAD)
    P.health = 4
    P:set_behavior(1)
    players[1].user_data.lives = players[1].user_data.lives - 1
    print(F"You have {players[1].user_data.lives} lives left!")
    return true
  end
end, ON.PRE_LOAD_SCREEN)
```

Malacath-92:
This doesn't seem to reset user data on a quick restart after dying

Malacath-92:
Oh ye, this is probably related to the comment right above mine 🤦‍♀️

PR #245, opened 2022-07-16
Iwyu

Malacath-92:
Add support for IWYU and Unity builds. Also apply all IWYU fixes

Things I should fix before merging:
- [x] build iwyu executables for windows (need to make a separate repo for that)
- [x] auto-download iwyu executables when enabled
- [x] make sure to run clang-format after a file was formatted


src/game_api/sound_manager.hpp
```diff
@@ -1,18 +1,24 @@
...
 #include <optional>
 #include <string>
 #include <string_view>
+#include <type_traits>
 #include <unordered_map>
+#include <utility>
 #include <variant>
 #include <vector>
 
+#include "D:/zollk/Documents/Projects/llvm-project/build/Release/lib/clang/13.0.1/include/stddef.h"
```
Mr-Auto, line 18: don't think this will work correctly (?)

src/game_api/string_hashes.cpp
```diff
@@ -1963,3 +1970,8 @@ const std::map<uint32_t, STRINGID> string_hashes = {
     {0x13aced89, 1959},
     {0x796b612f, 1960},
 };
+
+const std::map<uint32_t, STRINGID>& get_string_hashes()
+{
+    return string_hashes;
+}
```
Mr-Auto, line 1977: This is an auto generated file so would be better to make this change in the generator (also no empty line at the end)

src/game_api/texture.cpp
```diff
@@ -1,7 +1,18 @@
...
+#include <cstring>
+#include <functional>
+#include <list>
+#include <mutex>
+#include <new>
+#include <type_traits>
+#include <unordered_map>
+#include <utility>
+
+#include "D:/zollk/Documents/Projects/llvm-project/build/Release/lib/clang/13.0.1/include/stddef.h"
```
Mr-Auto, line 12: Again

Mr-Auto (changes requested):

src/game_api/sound_manager.hpp
Malacath-92: Yeah, idk why those came in there. How did you find them so quick  xD

Mr-Auto: well, most of the changes are adding/removing includes, so any problems would be catch by the compiler anyway, so while scrolling thru the changes, my eyes just catch that super long include line :smile:

src/game_api/string_hashes.cpp
Malacath-92: Wasn't aware that this was generated. Which file generates it?
I will add a comment to make sure that is more obvious in the future.

Mr-Auto: `strings_get_hashes.py`, we can also make it output to console, so you would just copy paste the array instead of generating the whole file

Malacath-92:
This is as good as it gets now. @Mr-Auto or @Dregu if you wanna try it simply `cmake .. -DIWYU=ON -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++` or whatever command you usually use to configure, just add `-DIWYU=ON` to it. Now every file should be compiled with clang and then with iwyu, giving you output for issues. You can also `cmake --build . --target=apply_iwyu_fixes` to apply fixes to all files or `cmake --build . --target=apply_iwyu_fixes_changes` to just your locally changed files (treating .cpp/.c, .hpp, .h and .inl files as tuples).
I build Windows binaries in [my fork of IWYU](https://github.com/Malacath-92/include-what-you-use) so if we need other binaries than 13.0.0 or 13.0.1 I can build them there.

Dregu (approved):
At least it didn't make anything worse and fixed ninja builds for me.

PR #247, opened 2022-07-28
Hook vanilla text rendering

Dregu:
They wanted an all death messages tracker and I made a dumb way to convert string pointers back to string ids. I don't know if this should even make it to the api.

Here's a simple tracker I made to test it.

```lua
first = hash_to_stringid(0x5a52a061)
last = hash_to_stringid(0x5c9b2332)
total = (last-first)/2
num = 0
all = {}
for id=first,last,2 do
  all[id] = { done=false, str=F"[{get_string(id)}] {get_string(id+1)}" }
end
out = F"{num}/{string.format('%d', total)}\n"
for k,v in pairs(all) do
  out = out .. v.str .. "\n"
end
set_callback(function(id)
  if id >= first and id <= last and not all[id].done then
    all[id].done = true
    num = 0
    for k,v in pairs(all) do if v.done then num = num + 1 end end
    --prinspect(id, get_string(id), F"{num}/{string.format('%d', total)}")
    out = F"{num}/{string.format('%d', total)}\n"
    for k,v in pairs(all) do
      if not v.done then
        out = out .. v.str .. "\n"
      end
    end
  end
end, ON.DEATH_MESSAGE)

set_callback(function(ctx)
  ctx:draw_text(-1, 0.78, 13, out, 0xffffffff)
end, ON.GUIFRAME)

set_callback(function()
  players[1].health = 1
end, ON.LEVEL)
```


src/game_api/strings.cpp
```diff
@@ -151,6 +151,17 @@ const char16_t* get_string(STRINGID string_id)
...
 
+STRINGID pointer_to_stringid(size_t ptr)
+{
+    auto strings_table = get_strings_table();
+    for (STRINGID i = 0; i < g_original_string_ids_end; ++i)
+    {
+        if ((size_t)strings_table[i] == ptr)
+            return i;
+    }
+    return g_original_string_ids_end;
```
Mr-Auto, line 162: This can be solved with simple math, just subtract the ptr from address of the stings_table and divide that by 8 (or something like that)

Mr-Auto (changes requested):

src/game_api/strings.cpp
Dregu: But the table is ** and what I have is the last *

Mr-Auto: oh wait, i dumb

Dregu: It's not actually always even in the strings table, when the string needed formatting for example. But it works for this very specific need of converting death messages to stringids...

PR #248, opened 2022-07-31
Add shader reload function

Malacath-92:


Malacath-92:
I abused this PR to add the fix to the gender setter since it was so small anyways.

PR #249, opened 2022-08-08
Particles

Malacath-92:
Add @jaythebusinessgoose's requests for particles


src/game_api/particles.hpp
```diff
@@ -39,7 +46,7 @@ struct ParticleDB
     uint8_t red;
     uint8_t green;
     uint8_t blue;
-    uint8_t unknown28;
+    uint8_t unkown28;
```
jaythebusinessgoose, line 49: This became spelled wrong.

src/game_api/rpc.cpp
```diff
@@ -1130,13 +1130,13 @@ void extinguish_particles(ParticleEmitterInfo* particle_emitter)
         static free_particleemitter_func* fpf = (free_particleemitter_func*)(offset);
         if (particle_emitter != nullptr)
         {
-            if (particle_emitter->unknown26 != 0)
```
jaythebusinessgoose, line 1133: I didn't actually test the changes I made to this file. I just made it compile in a way that looks like it should be correct.

Malacath-92: Tested and checked and could actually simplify this

PR #250, opened 2022-08-10
Add ui fonts for all supported game languages

Dregu:
These are random default Windows 10 or 7 fonts. I guess playlunky should copy this.

### DIFF
--- a/src/injected/ui.cpp
+++ b/src/injected/ui.cpp
@@ -6320,40 +6320,100 @@ void load_font()
     ImGuiIO& io = ImGui::GetIO();
     io.FontAllowUserScaling = false;
     PWSTR fontdir;
+
+    ImFontConfig font_config;
+    font_config.MergeMode = true;
+    font_config.EllipsisChar = u'\u2026';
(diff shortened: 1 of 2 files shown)


Malacath-92 (approved):

src/injected/ui.cpp
```diff
@@ -6320,40 +6320,100 @@ void load_font()
...
+    std::string font_jp;
+    std::string font_ko;
+    std::string font_ru;
+    std::string font_zhcn;
+    std::string font_zhtw;
+    std::string font_emoji;
+
     if (SHGetKnownFolderPath(FOLDERID_Fonts, 0, NULL, &fontdir) == S_OK)
     {
+        using cvt_type = std::codecvt_utf8<wchar_t>;
```
Malacath-92, line 6340: Just a tiny thing, you could be using `std::filesystem::path::string::c_str` instead of `std::codecvt` here.

Dregu: On the other hand this code was already here when I got there so I didn't have to write anything

Malacath-92: I just mentioned it bc it was nicer when stealing this code for PL  :P

PR #252, opened 2022-08-18
Add a method to fix impostor lake positions to match the positions of the entities.

jaythebusinessgoose:
When impostor lake entities are moved, it only moves the blue rendering. This change adds a method to fix the positions of impostor lake water physics boxes to match the positions of the entities.

### DIFF
--- a/src/game_api/state_structs.hpp
+++ b/src/game_api/state_structs.hpp
@@ -790,6 +790,15 @@ struct LiquidPool
     LiquidTileSpawnData tile_spawn_data;
 };
 
+struct LiquidLake
+{
+    uint32_t position1;
+    uint32_t position2;
+    uint32_t position3;
+    uint32_t lake_type;
+    Entity* impostor_lake;
+};
+
 struct LiquidPhysics
 {
     size_t unknown1; // MysteryLiquidPointer1 in plugin
@@ -815,6 +824,11 @@ struct LiquidPhysics
             LiquidTileSpawnData stagnant_lava_tile_spawn_data;
         };
     };
+    size_t unknown2;
+    size_t unknown3;
+    LiquidLake* start_lakes;
+    LiquidLake* end_lakes;
+    size_t unknown6;
 };
 
 struct AITarget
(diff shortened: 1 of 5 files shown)


src/game_api/state_structs.hpp
```diff
@@ -815,6 +824,11 @@ struct LiquidPhysics
             LiquidTileSpawnData stagnant_lava_tile_spawn_data;
         };
     };
+    size_t unknown2;
+    size_t unknown3;
+    LiquidLake* start_lakes;
+    LiquidLake* end_lakes;
+    size_t unknown6;
```
Mr-Auto, line 831: That's a vector

also can you add the the next one as well? `uint32 total_liquid_spawned;` it's "total number of spawned liquid entities, all types (not impostor's)"

Mr-Auto (changes requested):

src/game_api/state_structs.hpp
Malacath-92: A `game_vector` (using `malloc` and `free`) or a `custom_vector` (using their custom mem-heap)?

Mr-Auto: what's the difference?

it should only matter if you want to add/remove stuff from it, for just from it, it can be simply `std::vector`

jaythebusinessgoose: I updated this to be a std::vector, and also added the rest of the values from the plugin (even though most of them are unknowns).

Malacath-92: Yeah sure, you can just keep it a `std::vector` but one day when someone decides to use it they might not think about what this really is and break the game in subtle ways. So imho it's worth it to quickly check what this really is when adding it. You just check if it calls `malloc` or not during a `push_back`.

jaythebusinessgoose: I looked a bit, and I have no idea what it's doing to allocate more space. I think it's happening at around 229b7478, but I don't understand it.

jaythebusinessgoose: Updated to use custom_vector based on your recommendation of the alloc at the line I linked.

PR #253, opened 2022-08-27
Some stuff from the plugin, some comments, added music in `GameMenager` and some vtable work

Mr-Auto:


src/game_api/state_structs.hpp
```diff
@@ -976,3 +979,15 @@ struct ShopsInfo
...
+{
+    uint32_t* timer;
+    TextRenderingInfo** layout;
+    size_t unknown3;
+    size_t unknown4;
+    uint32_t unknown5;
+    float unknown6;
+    uint32_t unknown7;
+    uint32_t unknown8;
+};
```
Malacath-92, line 993: mission new-line

src/game_api/entities_chars.hpp
```diff
@@ -35,14 +36,15 @@ class Ai
     int32_t timer;
     /// AI state (patrol, sleep, attack, aggro...)
     int8_t state; // 8 = HH agro
-    int8_t unknown12;
+    int8_t last_state;
     /// Levels completed with, 0..3
     uint8_t trust;
     /// How many times master has violated us
```
Malacath-92, line 42: I just saw this on here. Can we please not use the word master.

Malacath-92 (changes requested):

src/game_api/entities_chars.hpp
Mr-Auto: wdym?

Dregu: Oh no bad word maybe slavery will come back if you repeat it three times. I don't care, but you should probably remove strings00 and the sound system too.

Malacath-92: If we have previously decided to censor the word damsel then I think not using master is just natural  ;)

Malacath-92: Your reference to the sound system and strings00.str is excessively childish. The usage there is in a clearly different context than what is used here. If you can't see that I am seriously sorry for you.

Dregu: This is all a joke to me like the thing I wrote in the docs a long time ago, cause the game is clearly joking on the subject too. Robots waiting for their old masters in the journal is indeed different vibe. This just sounds like virtue signalling. I guess all the human sacrifice and murder in the game is fine though, as long as some arbitrary words are cancelled every year. Do what you want, I don't care.

In my mind we changed damsel because the game calls them pets btw, nothing to do with censorship.

Malacath-92: It's an enigma to me who I would be virtue signaling to. Did you expect me to put this into the changelog or something?

But whatever, I will indeed do this somewhere else if I want to. Just because doing what I think is right publicly suddenly makes it bad I won't stop doing it.

src/game_api/search.cpp
```diff
@@ -1745,6 +1745,13 @@ std::unordered_map<std::string_view, AddressRule> g_address_rules{
             .at_exe()
             .function_start(),
     },
+    {
+        "play_sound"sv,
```
Mr-Auto, line 1749: This is a different than `construct_soundposition_ptr` and I'm not sure why, the game uses this function for all simple short sounds, and for everything that needs the position to be tracked it uses `construct_soundposition_ptr` and in most cases it's stored somewhere in entity memory. But both seam to make more or less the same thing, creating `SoundPosition` (now `SoundMeta`)

Mr-Auto: next issue, how to use the `VANILLA_SOUND` enum, as that contains strings and this uses ids, should I just dump the data to some vector/map when it's makes the enum or something?

src/game_api/script/usertypes/sound_lua.cpp
```diff
@@ -179,6 +179,11 @@ void register_usertypes(sol::state& lua, SoundManager* sound_manager)
         "playing",
         &SoundMeta::playing);
 
+    lua.new_usertype<BackGroundSound>(
+        "BackGroundSound",
```
Dregu, line 183: Since we're exposing it, it's Background, one word, no G.

src/game_api/script/usertypes/screen_lua.cpp
```diff
@@ -107,6 +107,10 @@ void register_usertypes(sol::state& lua)
         &ScreenTitle::particle_torchflame_flames_animated,
         "particle_torchflame_ash",
         &ScreenTitle::particle_torchflame_ash,
+        "background_music",
```
Dregu, line 110: Nice, but I'll just nitpick again that we now have a "backgroundmusic", "background_music" and "bg_music" exposed, and isn't all music background music, so maybe just call all of them "music"?

Mr-Auto: Like this? or you mean the struct name as well?

Dregu: The usertype names are fine.

src/game_api/state.hpp
```diff
@@ -51,25 +51,25 @@ struct StateMemory
     float fadevalue; // 0.0 = all visible; 1.0 = all black
     uint32_t fadeout;
     uint32_t fadein;
-    uint32_t loading_black_screen_timer; // if state.loading is 1, this timer counts down to 0 while the screen is black (used after Ouroboros, in
-                                         // credits, ...)
-    uint8_t ingame;
-    uint8_t playing;
+    uint32_t loading_black_screen_timer; // if state.loading is 1, this timer counts down to 0 while the screen is black (used after Ouroboros, in credits etc.)
+    /// True when you have control over your character
+    bool ingame;
```
Dregu, line 56: This is a breaking change unfortunately... Probably not widely used but it is used even in our examples.

Mr-Auto: i fixed it, but the autodoc ignores my `NoDoc` comment :/

Dregu: Right, I didn't add nodoc to the usertype parser cause it was such a hack already to get it working without rewriting the whole thing

src/game_api/script/usertypes/state_lua.cpp
```diff
@@ -242,8 +242,18 @@ void register_usertypes(sol::state& lua)
     statememory_type["screen_last"] = &StateMemory::screen_last;
     statememory_type["screen"] = &StateMemory::screen;
     statememory_type["screen_next"] = &StateMemory::screen_next;
-    statememory_type["ingame"] = &StateMemory::ingame;
-    statememory_type["playing"] = &StateMemory::playing;
+    statememory_type["in_game"] = &StateMemory::playing;
```
Dregu, line 245: oops

src/game_api/state.hpp
Dregu: If you want, you can also think of a better name for these new variables. They sound and behave almost exactly the same but I don't remember the difference.

Dregu: Or at least document it. I think ingame is everything after the menu, and playing is in an actual level.

Dregu: I added NoDoc support, merging soon

Dregu: `playing` actually seems like basically `state.screen == SCREEN.LEVEL`, having control of a character is misleading and you got them wrong way around. How about `in_game` and `in_level`

Mr-Auto: I added the comments, but i see some bugs with them now, like when going to options menu both are set to false, but going back only sets the ingame to true, and the playing stays false until next level

ingame is weird 
After the intro cutscene is set to true until you press the button to go to menu
When going to online lobby, is set to true, but not to false when going back, to set it to false again you can go to character select or Player profile and back (is set to false on the way back, not opening it)
Going to the seed input sets ingame to true, going back (correctly?) sets it to false
Is set to true in arena team select, after selecting team is set to false, then in arena itself to true again, the `playing` stays false

When going first time to the camp, only ingame is set to true, but when you die in a level and go to camp, both are true, even thou playing is set to false on the death screen

Dregu: Some of that weirdness almost seems like a bug in the game, or what even is the function of these variables

Mr-Auto: dunno, a added a break point with x64dbg and they are just written to a static memory, and the only access/write to it seams to be that code. Nothing else

Dregu: If it can't be trusted after going to options etc. seems kinda pointless to have at all and state.screen can be used for the same purpose. Maybe that's what the devs did too and forgot about this...

PR #255, opened 2022-08-29
Expose emitted_particles2 as the backlayer particles

jaythebusinessgoose:


src/game_api/rpc.cpp
```diff
@@ -1124,7 +1124,7 @@ void extinguish_particles(ParticleEmitterInfo* particle_emitter)
     if (particle_emitter != nullptr)
     {
         generic_free(particle_emitter->emitted_particles.memory);
```
jaythebusinessgoose, line 1126: Moving the particle count into the struct means this is now pointing to a different point in memory, so I'm once again not sure if this will still work or if it needs to be updated to free at the right spot.

Malacath-92: The memory pointer still points to the same address, nothing has changed, you're good.

jaythebusinessgoose:
> @jaythebusinessgoose I was just gonna merge this, but thought maybe it would be good to have a sample script for these. Do you have time to make one?

@Malacath-92 I added the mole particle stuff as an example script.

PR #259, opened 2022-09-20
Update

Mr-Auto:
Not stable


Dregu:
Doesn't seem to work without steam (so with goldberg) atm

Mr-Auto:
Works for me

Dregu:
Oh I had to reset the ini there, but not in my steam folder

PR #266, opened 2022-10-21
Add ability to mess with journal pages

Mr-Auto:


Dregu:
Just fyi, I have api for hooking get_achievement and set_achievement, checking for unlocked achievements and setting the hidden state to completely customize the existing Feats page, so don't do that maybe. I'll push a pr later.

Mr-Auto:
@Dregu don't do .. what exactly? how is this related?

Dregu:
I don't know what your plans were for the things you have not done yet. It's a page in the journal, it's pretty related...

Mr-Auto:
I've done all that i wanted, that's why i marked this PR as ready for review

Your PR is not related so much to the journal, but more to the back-end of the achievements. Goal of this PR was to find a way to add more pages to the "Feats" so that people can leave the "Feats" untouched and just add more pages. Then it grow into completely removing and adding pages at will, in the whole journal, not just "Feats". So people can use this to make custom pages for their custom entities, themes etc.

Dregu:
Yeah nice and the features work great together. I was just warning that maybe don't add exactly the same thing by accident when I'm asleep.

PR #270, opened 2022-11-02
Vanilla theme hooks

Dregu:
I did something. I'm pretty sure a lot of stuff is broken too.

```lua
-- disable theme functions
a = set_pre_theme(THEME.DWELLING, THEME_OVERRIDE.SPAWN_BORDER, function(theme)
  return true
end)

-- override theme functions
b = set_pre_theme(THEME.DWELLING, THEME_OVERRIDE.TEXTURE_DYNAMIC, function(theme, texture)
  if texture == DYNAMIC_TEXTURE.FLOOR and math.random() < 0.5 then
    return TEXTURE.DATA_TEXTURES_FLOOR_JUNGLE_0
  end
end)

-- fix things in post
c = set_post_theme(THEME.JUNGLE, THEME_OVERRIDE.SPAWN_EFFECTS, function(theme)
  if test_flag(state.presence_flags, 2) then
    state.camera.bounds_bottom = 56.5
  end
end)

--clear_theme_callback(THEME.DWELLING, a)
--clear_theme_callback(THEME.DWELLING, b)
--clear_theme_callback(THEME.JUNGLE, c)
```


Mr-Auto:
Didn't look much thru this, but i believe you don't need separate hook functions for void, bool, uint8, uint32, you can just always have it as returning size_t, then then throw away the upper bits and cast it depending on the type you need, or for bool, you can just use bitwise

The only exception is float which needs to have separate function

Dregu:
Yeah I kinda saw that too when it was done, but it was already done 🤷🏻. Or could've probably used some kind of templates, but I'm dumb.

Dregu:
You can always suggest your optimizations in a fork, I won't be refactoring this before next week anyway.

Dregu:
So there is a bug that all the functions of the same signature just run the same hook, cause that's just how our vtable_hook works I guess.

```c++
using VoidFun = void(ThemeInfo * self);
hook_vtable<VoidFun>(
    theme, [](ThemeInfo* self, VoidFun* original)
    {   DEBUG("THREE");
        original(self); },
    0x3);
hook_vtable<VoidFun>(
    theme, [](ThemeInfo* self, VoidFun* original)
    {   DEBUG("FOUR");
        original(self); },
    0x4);
hook_vtable<VoidFun>(
    theme, [](ThemeInfo* self, VoidFun* original)
    {   DEBUG("FIVE");
        original(self); },
    0x5);

/* prints:
[debug] FIVE
[debug] FIVE
[debug] FIVE */

/* stupid workaround */

using VoidFun = void(ThemeInfo * self);
using VoidFun2 = void(ThemeInfo * self, void* random);
using VoidFun3 = void(ThemeInfo * self, bool crap);
hook_vtable<VoidFun>(
    theme, [](ThemeInfo* self, VoidFun* original)
    {   DEBUG("THREE");
        original(self); },
    0x3);
hook_vtable<VoidFun2>(
    theme, [](ThemeInfo* self, void* random, VoidFun2* original)
    {   DEBUG("FOUR");
        original(self, random); },
    0x4);
hook_vtable<VoidFun3>(
    theme, [](ThemeInfo* self, bool crap, VoidFun3* original)
    {   DEBUG("FIVE");
        original(self, crap); },
    0x5);

/* prints:
[debug] THREE
[debug] FOUR
[debug] FIVE */
```

Dregu:
Well I've spent days trying to figure out how to fix this hook so that you could define the vtable index at runtime and have multiple hooks with the same signature. This feature is scrapped.

Malacath-92:
I can fix that too once I'm at a PC, but you know that's a very volatile expectation for me these days. So just gotta have patience.

Malacath-92:
Now I see your issue, for the first time these are runtime values for the vtables. So it's not as trivial as I thought. But we can just put a stupid wrapper around it that should be a jump table when compiled.

Malacath-92:
No time to test, you can give this a try

Dregu:
Oh the stupid wrapper looks exactly like the stupid thing we would run into having to define the index at runtime 50 times when doing basically the same thing.

Dregu:
It seems to work. At least Jays example that sparked this problem is fixed. I don't know where I went wrong with the template argument stuff, everything you did looks pretty much what I tried to do, sans the stupid wrapper.

Dregu:
#284

PR #271, opened 2022-11-03
Expose+rename play_warning_sound→trigger_explosion

mauvealerts:


src/game_api/entities_items.hpp
```diff
@@ -31,8 +31,8 @@ class Backpack : public Movable
     virtual void v95() = 0; // just return
     virtual void on_putting_on(Entity* who) = 0;
     virtual void on_putting_off(Entity* who) = 0;
-    virtual bool is_active() = 0;          // for jetpack returns jetpack.flame_on, for capes Cape.floating_down, for hoverpack, hoverpack.is_on, teleporter and powerpack return false
-    virtual void play_warning_sound() = 0; // plays the warning sound before explosion for jet, teleporter, hover and power pack
+    virtual bool is_active() = 0;         // for jetpack returns jetpack.flame_on, for capes Cape.floating_down, for hoverpack, hoverpack.is_on, teleporter and powerpack return false
+    virtual void trigger_explosion() = 0; // Causes the backpack to play its warning sound and triggers the explosion
```
Mr-Auto, line 35: Not really a fan of the name, there is `explosion_trigger` variable, now the `trigger_explosion` function, feels too similar

mauvealerts:
I think there's also downsides to it being too dissimilar. When I first
picked a name, it was "start_detonation" because that seemed accurate.
However, it seemed weird to have it so different from the 'explosion'
variables. "start_explosion" also seems pretty weird and
"prepare_explosion" is pretty long …  Overall, I'm ambivalent

This comment is exactly why I tried to have the bike-shedding discussion
(in #s2-modding-tooldevs) before I made this PR.

For clarity: If you want a different name, please suggest one. I'd don't
have a strong preference, nor is there an alternative with broad support.

Dregu:
I think the name is good and the similarity tells the user they are related. The comment perfectly describes the difference. 7/5 with rice.

PR #272, opened 2022-11-12
Options

Dregu:
- Reflect programmatic changes to `options` table in the gui/backend
- Add missing default value for `register_option_combo`
- Add `register_option_callback` to draw fancy options with window api
- Add `win_section` and `win_indent`
- Move `options` to script env, I guess it was always global, lol

### DIFF
--- a/docs/src/includes/_globals.md
+++ b/docs/src/includes/_globals.md
@@ -81,7 +81,7 @@ Provides a read-only access to the save data, updated as soon as something chang
 
 > Search script examples for [options](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=options)
 
-Table of options set in the UI, added with the [register_option_functions](#register_option_int).
+Table of options set in the UI, added with the [register_option_functions](#Option-functions). You can also write your own options in here or override values defined in the register functions/UI before or after they are registered. Check the examples for many different use cases and saving options to disk.
 ### prng
 
 #### [PRNG](#PRNG) prng
--- a/docs/src/includes/_globals.md
+++ b/docs/src/includes/_globals.md
@@ -152,9 +152,10 @@ Clears a previously set callback
 
 > Search script examples for [register_option_callback](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=register_option_callback)
 
-#### nil register_option_callback(string name, function callback)
+#### nil register_option_callback(string name, object value, function on_render)
 
-Add custom options using the window drawing functions. Your callback will be called with a [GuiDrawContext](#GuiDrawContext) as a parameter and everything drawn in it will be rendered in the options window and the return value saved to `options[name]`.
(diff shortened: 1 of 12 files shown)


docs/src/includes/_globals.md
```diff
@@ -147,6 +147,15 @@ Clears a callback that is specific to a screen.
...
 Clears a previously set callback
 
+### register_option_callback
+
+
+> Search script examples for [register_option_callback](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=register_option_callback)
+
+#### nil register_option_callback(string name, function callback)
+
+Add custom options using the window drawing functions. Your callback will be called with a [GuiDrawContext](#GuiDrawContext) as a parameter and everything drawn in it will be rendered in the options window and the return value saved to `options[name]`.
```
Mr-Auto, line 157: Can you format it like i did in my PR "The callback signature is" in separate line?

Dregu: Yep

src/game_api/script/usertypes/socket_lua.cpp
```diff
@@ -94,5 +196,29 @@ void register_usertypes(sol::state& lua)
...
+            luaL_error(lua, err.c_str());
+        }
+        return std::nullopt;
+    };
+
+    /// Send an asynchronous HTTP GET request and run the callback when done. If there is an error, response will be nil and vice versa.
+    /// The callback signature is nil on_data(string response, string error)
+    lua["http_get_async"] = [](std::string url, sol::function on_data) -> HttpRequest*
+    {
+        HttpRequest* req = new HttpRequest(std::move(url), std::move(on_data));
```
Malacath-92, line 220: This is gonna cause race conditions on Lua, isn't it?

Malacath-92 (changes requested):

src/game_api/script/usertypes/socket_lua.cpp
Dregu: Don't know, that's why I have tried to keep away from thread stuff.

Dregu: This is the sort of thing that I'm going to need an example and a solution for or I'll just remove the thing cause you said it won't work.

Malacath-92: Anything that uses `get_calling_backend` and family has a handle to a LuaBackend that is protected by a mutex. Any access to the Lua vm through any other handle is not threadsafe. I don't know the exact names of stuff off the top of my head, but you should be able to compare to anything else that uses callbacks. So the approach would be to `get_calling_backen_id` here, then pass a lambda capturing said id and the cb, inside of which you `get_backend_from_id` or smth and then `handle_function_woth_return<void>` or smth like that. Then you can also store a std:: function in your HttpRequest`. Ig that whole process could be wrapped easily, I'll take a look at it next time I sit at a PC.

Dregu: Thanks. Yeah I kinda just copied the (probably also broken) udp function I made ages ago and it slipped though the radar then. I'll make another branch for fixing the net stuff and merge options in the meantime. I just did http on a whim cause someone mentioned it.

PR #273, opened 2022-11-12
Stuff

Mr-Auto:
- Add some examples
- Standardize the "The callback signature is" comment for callbacks and add the missing ones
- Remove code blocks from the parameters for ON enum
- fix `info_dump`
- fix links in `entities-hierarchy.md`


Dregu:
Nice, some boring things off my todo list. I was also going to give `ON` its own toplevel section called Events, styled like the functions and not in a tight table, supporting individual examples from docs/examples.

docs/examples/players.lua
```diff
@@ -0,0 +1,3 @@
+> Make the player invisible, use only in single player only mods
```
Dregu, line 1: I think this (and all the others) are supposed to be --lua comments and not markdown.

Dregu: You can use > and multiple code blocks in .md examples to separate different examples better, or use one .lua example that has to be all lua. But I guess this is just a brainfart.

PR #274, opened 2022-11-19
emmylua - callback signatures, some fixes

estebanfer:


docs/game_data/spel2.lua
```diff
@@ -73,36 +73,36 @@ function prinspect(...) end
 function messpect(...) end
 ---Adds a command that can be used in the console.
 ---@param name string
----@param cmd fun(): any
+---@param cmd function
 ---@return nil
 function register_console_command(name, cmd) end
 ---Returns unique id for the callback to be used in [clear_callback](#clear_callback). You can also return `false` from your function to clear the callback.
 ---Add per level callback function to be called every `frames` engine frames. Timer is paused on pause and cleared on level transition.
----@param cb fun(): any
```
Malacath-92, line 81: Maybe we should add the ones that are now missing in the comments?

docs/game_data/spel2.lua
```diff
@@ -1321,8 +1321,8 @@ function get_short_tile_code_definition(short_tile_code) end
 ---Set `is_valid` to `nil` in order to use the default rule (aka. on top of floor and not obstructed).
 ---If a user disables your script but still uses your level mod nothing will be spawned in place of your procedural spawn.
 ---@param procedural_spawn string
----@param do_spawn fun(): any
----@param is_valid fun(): any
+---@param do_spawn function
```
Malacath-92, line 1324: And adjust the comments that are not caught by your new logic?

estebanfer: I probably would have to add something different when there's multiple callbacks in a single function.

docs/game_data/spel2.lua
estebanfer: Agree

estebanfer:
I also could add callback signatures to classes like CustomMovableBehavior, but the comments there didn't have the parameter type

Malacath-92:
We can add them there too, shouldn't be much trouble. I can add some comments on main if you like (when I have time)

Mr-Auto:
While this PR is still up: could you add that the `<br/>` in the comments is omitted for the spel2.lua?
could also do something with `[name](#name)` links, but it's not that big of a deal

Mr-Auto:
Also i added new constructor for AABB in my PR #276 (from tuple so it's only for the internal use), but the `/// NoDoc` didn't work for the emmylua so spel2.lua got that constructor

estebanfer:
Should I make deprecated functions to also be printed on spel2.lua and mark them as [---@deprecated](https://github.com/sumneko/lua-language-server/wiki/Annotations#deprecated)?

Malacath-92:
> Should I make deprecated functions to also be printed on spel2.lua and mark them as [---@deprecated](https://github.com/sumneko/lua-language-server/wiki/Annotations#deprecated)?

I'd avoid them there.

estebanfer:
You sure? I think it would be better to include the deprecated ones, so when people update spel2.lua they still see the functions like set_post_statemachine instead of just showing that the function doesn't exist, and also showing the warning that the function is deprecated, like this:
image

Mr-Auto:
> You sure? I think it would be better to include the deprecated ones, so when people update spel2.lua they still see the functions like set_post_statemachine instead of just showing that the function doesn't exist, and also showing the warning that the function is deprecated, like this: image](https://user-images.githubusercontent.com/20376110/214326668-37c4b8b8-8a0f-47b4-bd0a-56c18ba7a4a0.png)

I think the main reason not to include them is autocomplete, very little people will probably read that message, and the whole idea of deprecated functions is to discourage people from using them. 

I also think in the API docs they should be much more clearly marked as well

gmjosack:
I end up with functions that just aren't in the spel2 lua all the time and just assume they're missing rather than deprecated. If a deprecation does a strike through like in estabanfer's example thats much more clear that it shouldn't be used than the yellow squiggle which could just mean it's not defined to to a bug or using an old spel2.lua or any number of other reasons

estebanfer:
I agree with gary, had the same thing happen before with functions showing as missing.
Also, on autocomplete for vscode, deprecated functions show like this
image

estebanfer:
I think that I'm done with the changes, though I think I need to change the constructors to use `---@overload` because on autocomplete it only shows the last constructor declared instead of all of them, but I think I'll leave that for another PR.
Also ig I'll leave the deprecated functions stuff it for another PR if it's going to be added.

PR #276, opened 2022-11-22
Stuff

Mr-Auto:
- fix link to `entity hierarchy list`
- improve `entity_info` script by tiding it to the camera instead of player
- remove some alias functions
- add `static` keyword to most of the `get_address` uses
- deprecate `set_blood_multiplication` as it doesn't work (probably never did)
- make `set_on_damage` available only for movable entities
- add missing events to `ON` enum
- fix `get_io` in the doc
- make sure the `The callback signature is` is always in new line
- replace `DEFINE_ACCESSOR` with template function
- add `GuiDrawContext` parameter to the window widget
- add split function for the `AABB` and fix it in the doc for `Vec2`


src/game_api/character_def.cpp
```diff
@@ -41,7 +41,7 @@ std::span<CharacterDB> GetCharacterDefinitions()
 {
     static std::span<CharacterDB> static_character_table = []()
     {
-        const auto character_db = get_address("character_db");
+        static const auto character_db = get_address("character_db");
```
Malacath-92, line 44: That static is not helping, the lambda is already invoked only a single time.

src/game_api/memory.hpp
```diff
@@ -119,23 +119,11 @@ requires std::is_trivially_copyable_v<T> void write_mem_prot(void* addr, const T
...
-DEFINE_ACCESSOR(u32, uint32_t);
-
-DEFINE_ACCESSOR(u64, uint64_t);
-
-DEFINE_ACCESSOR(i32, int32_t);
-
-DEFINE_ACCESSOR(i64, int64_t);
-
-DEFINE_ACCESSOR(f32, float);
+template <typename T>
```
Malacath-92, line 122: 🙌🙌

src/game_api/script/lua_backend.hpp
```diff
@@ -46,6 +46,7 @@ enum class ON
     TITLE = 3,
     MENU = 4,
     OPTIONS = 5,
+    // PLAYER_PROFILE = 6,
```
Malacath-92, line 49: why commented?

Malacath-92 (changes requested):

src/game_api/script/lua_backend.hpp
Mr-Auto: I assume they work, but haven't tested yet

src/game_api/memory.hpp
Mr-Auto: ?

Malacath-92: I just like this

PR #284, opened 2022-12-08
Crazy theme hooks

Dregu:
Made theme hooks with #275

Everything seems to work, made a kinda funny get_aux_id though.

```lua
-- change dwelling floor blue
state.level_gen.themes[THEME.DWELLING]:set_pre_texture_dynamic(function(theme, id)
  if id == DYNAMIC_TEXTURE.FLOOR then
    return TEXTURE.DATA_TEXTURES_FLOOR_ICE_0
  end
end)

-- disable path gen in jungle. also disables exit btw, we should really think about these names a bit
state.level_gen.themes[THEME.JUNGLE]:set_pre_virtual(THEME_OVERRIDE.GENERATE_PATH, function(theme)
  return true
end)

-- this shotgun shoots bubbles when the user is wet
  ent:set_pre_trigger_action(function(self, user)
    if user and user.wet_effect_timer > 0 then
      local x, y, l = get_position(self.uid)
      local vx = 0
      if test_flag(self.flags, ENT_FLAG.FACING_LEFT) then
        x = x - 0.5
        vx = -0.1
      else
        x = x + 0.5
        vx = 0.1
      end
      spawn_critical(ENT_TYPE.ACTIVEFLOOR_BUBBLE_PLATFORM, x, y, l, vx, 0)
      return true
    end
  end)
end, SPAWN_TYPE.ANY, MASK.ITEM, ENT_TYPE.ITEM_SHOTGUN)
```


Dregu:
Yeah I just managed to make more commits, fuck this

Malacath-92:
How about:
- new branch from main
- cherry-pick from this branch
- new PR and close this one

As it stands you somehoe even reverted some of my changes. I can try to do this when I'm at work.

Dregu:
Yeah I can do that and just copypaste from the last sane commit, there's like 4 blocks of code to copy.

PR #286, opened 2022-12-12
Yet another theme hooks pr

Dregu:
You didn't see anything.

Things are mostly there already, but we should think about the names really hard before merging at least. It would be preferable to keep most of the existing THEME_OVERRIDE names, if they are not totally misleading atm. And then there's the autodoc to fix.

```lua
-- change dwelling floor blue
state.level_gen.themes[THEME.DWELLING]:set_pre_texture_dynamic(function(theme, id)
  if id == DYNAMIC_TEXTURE.FLOOR then
    return TEXTURE.DATA_TEXTURES_FLOOR_ICE_0
  end
end)

-- disable path gen in jungle. also disables exit btw, we should really think about these names a bit
state.level_gen.themes[THEME.JUNGLE]:set_pre_virtual(THEME_OVERRIDE.GENERATE_PATH, function(theme)
  return true
end)

-- this shotgun shoots bubbles when the user is wet
  ent:set_pre_trigger_action(function(self, user)
    if user and user.wet_effect_timer > 0 then
      local x, y, l = get_position(self.uid)
      local vx = 0
      if test_flag(self.flags, ENT_FLAG.FACING_LEFT) then
        x = x - 0.5
        vx = -0.1
      else
        x = x + 0.5
        vx = 0.1
      end
      spawn_critical(ENT_TYPE.ACTIVEFLOOR_BUBBLE_PLATFORM, x, y, l, vx, 0)
      return true
    end
  end)
end, SPAWN_TYPE.ANY, MASK.ITEM, ENT_TYPE.ITEM_SHOTGUN)
```


Dregu:
There is still the problem that some of these are really not compatible with the old hooks that mess with theme vtables. I don't know what to do about that.

Malacath-92:
Yes, I still need to adjust something about the hooks we do in level_api.cpp but I haven't really thought of what yet. Afaik at least one of them could just be an observer, but that's more tricky with others. I thought we could hook via our new hooks, iff they are available and otherwise use the old stuff

Dregu:
I'll merge this tomorrow or something if nothing comes up.

PR #287, opened 2022-12-13
Random UI tweaks mostly

Dregu:
Most controversial thing is making `savegame` writable. But guess what: it was always mostly writable and and nothing bad happened. We have so many other ways to protect the save anyway in the tools.


docs/src/includes/_enums.md
```diff
@@ -638,6 +638,22 @@ Name | Data | Description
...
+
+Used in state.pause
+
+Name | Data | Description
+---- | ---- | -----------
+[MENU](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE.MENU) | 0x01 | Menu: Pauses the level timer and engine. Can't set, controller by the menu.<br/>
+[FADE](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE.FADE) | 0x02 | Fade/Loading: Pauses all timers and engine.<br/>
+[CUTSCENE](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE.CUTSCENE) | 0x04 | Cutscene: Pauses total/level time but not engine. Used by boss cutscenes.<br/>FLAG4:<br/>Unknown purpose: Pauses total/level time and engine. Might change this later!<br/>FLAG5:<br/>Unknown purpose: Pauses total/level time and engine. Might change this later!<br/>
+[FLAG4](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE.FLAG4) | 0x08 | 
+[FLAG5](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE.FLAG5) | 0x10 |
```
Mr-Auto, line 654: Something didn't work there with the comments?
the `FLAG5` is useful as it pauses everything but leaves one of the frame counter so set_global_interval works still (you seams to remove that info on the variable so maybe add it here?)

docs/src/includes/_enums.md
```diff
@@ -652,6 +668,22 @@ Name | Data | Description
...
+
+Name | Data | Description
+---- | ---- | -----------
+[MENU](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.MENU) | 1 | 
+[FADE](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.FADE) | 2 | 
+[CUTSCENE](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.CUTSCENE) | 3 | 
+[FLAG4](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.FLAG4) | 4 | 
+[FLAG5](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.FLAG5) | 5 | 
+[ANKH](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=PAUSE_FLAG.ANKH) | 6 | 
+
```
Mr-Auto, line 686: Are we really going to double every enum?

Dregu: I don't really like it either but people have demonstrated that they don't understand bits and we started numbering these like this for some reason ages ago. Have a better idea? At least we'd need a new function to set and clear these values with a bitmask, I don't even want to do that |& crap.

Mr-Auto: Technically you can make function in lua that will return the "index/flag number", not sure which one is better as both just complicate this stuff even more in my opinion, i can imagine someone just being more confused seeing two enum for one thing

Dregu: People were getting confused because "there's no way" to get the flag number from these bitmasks. Whatever, I removed PAUSE_FLAG and added a function to set bitmasks and you just have to know what you are doing.

Dregu:
The hell is the CI tripping on that now, I didn't touch it.

PR #288, opened 2022-12-14
Name a bunch of properties of entities

jaythebusinessgoose:


src/game_api/level_api.hpp
```diff
@@ -498,8 +498,8 @@ struct LevelGenSystem
         };
     };
     uint8_t flags;
-    uint8_t unknown39; // also flags?
-    uint8_t unknown40; // also flags?
+    uint8_t flags2;
+    uint8_t flags3;
```
Dregu, line 502: Is it just one 32bit flags? Or maybe we have just combined multiple 8bit flags to 32bit flags before, but it doesn't really matter...

Dregu: Oh I assumed there was just automatic padding added by aligning the next thing and there's actually 32 bits there.

src/game_api/entities_chars.hpp
```diff
@@ -102,11 +102,11 @@ class Player : public PowerupCapable
...
     int32_t i16c;
     float y_pos; // not sure why, seams to be the same as abs_y
-    uint8_t jump_flags_a;
+    uint8_t jump_lock_timer;
     /// can jump while airborne if greater than 0
     int8_t coyote_timer;
-    uint8_t jump_flags_c;
-    uint8_t jump_flags_d;
+    uint8_t hired_hand_name;
+    uint8_t swim_timer;
```
Mr-Auto, line 109: Some comment's maybe? like `hired_hand_name` is that a stringid? those usually come as 32bit value

src/game_api/entity_db.hpp
```diff
@@ -55,8 +55,8 @@ struct EntityDB
         };
         CollisionInfo default_collision_info;
     };
-    int32_t field_3C;
-    int32_t field_40;
+    int32_t collision2_mask;
+    int32_t collision_mask;
```
Mr-Auto, line 59: again, no comments/explanation/enum?

src/game_api/entity_db.hpp
```diff
@@ -86,7 +86,7 @@ struct EntityDB
     int32_t tile_y;
     uint8_t damage;
     uint8_t life;
-    uint8_t field_96;
+    uint8_t sacrifice_value;
```
Mr-Auto, line 89: Comment? is that a amount of favor you get for sacrifice? (last time i checked that was hardcoded)

src/game_api/level_api.hpp
Mr-Auto: I would still use 32bit, it doesn't do any hard really, unless you know the specifics of each flags variable

src/game_api/script/usertypes/entities_items_lua.cpp
```diff
@@ -819,6 +820,21 @@ void register_usertypes(sol::state& lua)
...
+        &OlmecShip::door_fx,
+        "smoke",
+        &OlmecShip::smoke,
+        "flight_time",
+        &OlmecShip::flight_time,
+        "has_spawned_jetflames",
+        &OlmecShip::has_spawned_jetflames,
+        sol::base_classes,
+        sol::bases<Entity, Movable>());
+
```
Mr-Auto, line 837: The proper way of adding new entity types is: add them in plugin: run `entity_casting.py` (it will generate new `entities-hierarchy.md`) then copy the output of the console to `entity_casting_lua.cpp` and `lua_console.cpp`

Mr-Auto (changes requested):

src/game_api/entity_db.hpp
```diff
@@ -55,7 +55,9 @@ struct EntityDB
         };
         CollisionInfo default_collision_info;
     };
+    // MASK, will only call collision2 when colliding with entities that match this mask.
```
Dregu, line 58: You dropped some slashes on the way here.

src/game_api/entity_db.hpp
```diff
@@ -86,7 +88,7 @@ struct EntityDB
     int32_t tile_y;
     uint8_t damage;
     uint8_t life;
-    uint8_t sacrifice_value;
+    uint8_t sacrifice_value; // Favor for sacrificing alive. Halved when dead (health == 0).
```
Dregu, line 91: And I believe the autodoc only considers triple comments on the previous lines.

src/game_api/script/usertypes/entities_items_lua.cpp
jaythebusinessgoose: Ok, I think I may have done it right.

src/game_api/entity_db.hpp
jaythebusinessgoose: didn't realize it was supposed to be 3, should be good

src/game_api/entity_db.hpp
jaythebusinessgoose: ah ok cool.

src/game_api/level_api.hpp
jaythebusinessgoose: I don't understand how it could be 32 bit when there are only 24 bits here.

I do know the specifics of each flag variable, I added the details to the plugin, but not sure how to name those for OL.

src/game_api/level_api.hpp
```diff
@@ -170,8 +170,8 @@ struct DoorCoords
 class ThemeInfo
 {
   public:
-    bool unknown1; // gets set to false for the jungle and temple theme in the cosmic ocean
-    bool unknown2;
+    bool allow_beehive;
+    bool allow_leprechaun;
```
Mr-Auto, line 174: Aren't leprechaun's unique to the "voice echoes" levels? you sure this is specific to the leprechaun alone?

PR #289, opened 2022-12-23
More vanilla draw stuff

Mr-Auto:
TODO:
- ~Make equivalents for the draw in world coordinates~
- ~Make some examples~


src/game_api/script/usertypes/vanilla_render_lua.hpp
```diff
@@ -46,6 +49,29 @@ class VanillaRenderContext
...
+    /// Use in combination with ON.RENDER_✱_HUD/PAUSE_MENU/JOURNAL_PAGE events
+    void draw_screen_triangle(const Triangle& triangle, Color color, bool filled, float thickness);
+
+    /// Draws a line on screen using the built-in renderer from point `A` to point `B`.
+    /// Use in combination with ON.RENDER_✱_HUD/PAUSE_MENU/JOURNAL_PAGE events
+    void draw_screen_line(const Vec2& A, const Vec2& B, Color color, float thickness);
+
+    /// Draw a line on screen from points using the built-in renderer
+    /// Draws from the first to the last point, use `closed` to connect first and last as well
+    void draw_screen_line(std::vector<Vec2> points, Color color, float thickness, bool closed);
```
Dregu, line 73: The gui equivalent is `draw_poly(points, thickness, color)`, we might want to keep the naming and argument order less confusing for all of these...

Dregu: I should also add `closed` to the end of those.

Mr-Auto: isn't it that `draw_poly` is meant to draw an actual shape? (filled and not as they tend to have two separate functions for that)
I could rename it, but then the `closed` parameter doesn't fit, as it's not really a polygon anymore when it is not closed.

So instead of changing this, why i don't just add `draw_poly` instead?

Dregu: `draw_poly` is for polyline, from the imgui function AddPolyline. We also have `draw_poly_filled` for filled convex polygons. These are also in the documentation so I don't know why I'm explaining them here...

Mr-Auto: I added `draw_polyline` and `draw_poly`, is that ok?

Dregu: You didn't address any of the points I made: naming and argument order is still inconsistent with the existing https://spelunky-fyi.github.io/overlunky/#GuiDrawContext functions draw_poly and draw_poly_filled... Now draw_poly is a line in gui context and filled polygon in this context...

Mr-Auto: It makes no sense to make them argument by argument identical. 
Also in my opinion GUI has way too many variants of the same function for no reason

Dregu: Ok but is there a reason to make everything different on purpose? This shit is impossible to learn.

Mr-Auto: That's why we have the API page? there is no equivalent to draw_image in the vanilla and no one seams to be confused about it

Dregu: Why are you fighting this? You just want to make the API confusing on purpose because you can always look up the stupid decisions in the docs?

Say you wanted to convert your old gui draw code to vanilla draw code.. Some things could be as easy as changing the ctx, but no because the devs were retards you need to rewrite everything!

PR #290, opened 2023-01-04
Only set camera bounds when toggling

mauvealerts:
The game actively manages `camera->bounds_bottom` in some levels, most obviously Hundun. #287 included a change that set the bounds on every frame (in commit `cbe954f`). Since the `bounds_bottom` it sets assumes a typical level, and it's called after the game sets `bounds_bottom`, the result is the bounds being "too low".

To address this problem, this PR removes setting the bounds on every frame. I think this re-introduces jank when the game is adjusting the bounds, e.g. you can move the camera down and it'll be pulled back up if the game sets `bounds_bottom`.

I considered only setting `bounds_bottom` if it was `-FLT_MAX`, but that seems really weird. I haven't thought of any other options.

### DIFF
--- a/src/game_api/search.cpp
+++ b/src/game_api/search.cpp
@@ -375,13 +375,13 @@ class PatternCommandBuffer
                 break;
             case CommandType::DecodePC:
                 offset = std::apply([=](auto... args)
-                { return ::decode_pc(exe, offset, args...); },
-                    data.decode_pc_args.as_tuple());
+                                    { return ::decode_pc(exe, offset, args...); },
+                                    data.decode_pc_args.as_tuple());
                 break;
             case CommandType::DecodeIMM:
                 offset = std::apply([=](auto... args)
-                { return ::decode_imm(exe, offset, args...); },
-                    data.decode_imm_args.as_tuple());
+                                    { return ::decode_imm(exe, offset, args...); },
+                                    data.decode_imm_args.as_tuple());
                 break;
             case CommandType::DecodeCall:
                 offset = mem.decode_call(offset);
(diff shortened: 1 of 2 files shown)


Dregu:
Removing this altogether breaks a lot more things than it fixes. Yes, setting the bottom bounds to wrongly calculated defaults was definitely wrong in a few levels, but at least the infinite bounds have to be enforced every frame exactly because hundun sets them every frame and also because it's a global option that should stay that way, and loading a new level sets default bounds again. Not enforcing it just desyncs the checkbox from the current state and it has to be toggled twice. That was probably the bug I was fixing by adding this line...

These changes might actually fix everything:
- enforce infinite bounds every frame when the "default bounds" option is off
- set calculated defaults once when the option is toggled on, focus is returned to player etc
  + check for state->logic->black_market for smaller bottom bounds

mauvealerts:
Yes, this is a partial revert of cbe954f. I agree that the behavior was
busted before that commit, and is busted with this PR in its current state.

I did try setting default bounds only occasionally, but concluded that
there were a bunch of cases to handle. I'll take a stab at a larger change
to better support the disabled/infinite bounds state, and try to think more
deeply about expected behavior…

I might look into a better version of computing default bounds, but my
impression is that isn't the goal of the feature. Maybe it's worth
re-labeling the checkbox to emphasize the removed/infinite bounds.

Dregu:
- Setting camera bounds for a level is part of a larger ThemeInfo function, so we need our own implementation simply to re-enable them
- There's never a reason to continuously force anything except the infinite bounds
- Flipping the options around (so default to "[ ] Disable camera bounds") might explain it better, I dunno
- I already pushed a simple fix for the real issues, can think about better naming later

mauvealerts:
Sounds great, thanks!

I don't have any concrete ideas beyond what you already pushed.

PR #297, opened 2023-01-27
Round two on TooManySkins

Dregu:
- Replace the actual vanilla texture definitions instead of setting entity textures to custom definitions so..
- ...we now automatically support all screen textures, character selection textures, ropes, posters, transitions...
- Now supports 4 players in local coop, with player1 in command of opening/closing
- Added `replace_texture(vanilla_id, texture_id)`
- Extracts heart color from the custom textures indicator, I dunno seemed neat


Dregu:
The playlunky Vfs is too confusing for me to make this work properly with the dds cache and convert new images on the fly on refresh, so I don't think I'll be doing that side. Cache and working with playlunky are pretty mandatory features though.

I did hack something together quickly to test this loading as a bugfix automatically, but I don't know what I'm doing. It worked with a few full sheets installed and crashed with a vanilla sheet 🤷🏻 https://github.com/spelunky-fyi/Playlunky/compare/main...Dregu:Playlunky:skinstuff

Dregu:
And the loose file loading shouldn't depend on preloading all the images anyway, it should just load any image files on the fly but also cache them for later.

src/game_api/render_api.cpp
```diff
@@ -153,6 +153,7 @@ void render_journal_page_place(JournalPage* page)
 }
 void render_journal_page_people(JournalPage* page)
 {
+    trigger_vanilla_render_journal_page_callbacks(ON::RENDER_PRE_JOURNAL_PAGE, JournalPageType::People, page);
```
Mr-Auto, line 156: WIP or you missed that it's only for this one page?

Dregu: I was too perplexed both by it not existing yes and being so easy to add I forgot to do the rest.

Mr-Auto: I think the main idea was: you can always draw on top of it with POST and most pages you can just make the elements invisible. 
Then I also added option to just make a blank page. So probably not many people had a need for it, but i don't hate it

PR #299, opened 2023-01-30
Move ON.SAVE callbacks to run last.

gmjosack:
Currently if you call save_script or save_progress from a callback that runs after ON.SAVE it will update your last_save time but not actually run the ON.SAVE callback.

This moves ON.SAVE callbacks into their own vector that always run last so manual saves work from any other callbacks.


gmjosack:
I don't currently have an OL environment setup so I haven't tested this but wanted to get it up for feedback if it's the right approach.

also noticed we weren't clearing load_callbacks in clear_all_callbacks so I added that as well

PR #301, opened 2023-02-08
Stuff

Mr-Auto:


src/game_api/script/lua_vm.cpp
```diff
@@ -97,6 +97,103 @@
...
+    }
+    auto end()
+    {
+        return p.end();
+    }
+
+  private:
+    void update()
+    {
+        StateMemory* main_state = State::get().ptr_main();
```
Dregu, line 139: I think it should use ptr_local, but yeah a very good change. We should probably make `state` to always return ptr_local too. And I don't know any simpler way to make the globals behave like this, maybe we should never have added any of them and just use functions.

Mr-Auto: A global variable doesn't need that much code, the issue is the containers as they have the `#[name]`, `pairs` and `ipairs` functions stuff

i used `ptr_main` as that's what it is right now, but can change it and the state if you think it's good idea

Dregu: Well setting it to local globally wouldn't make sense, but if it can always get the right thread, it probably should. It will be main usually anyway.

Dregu: All the get_entity functions should use layer_local too unless it causes any issues, I just found out they aren't...

Mr-Auto: calling `ptr_local` very already on crashes, i assume it's because it assumes many things to be a valid address, so it tries to access address 0 at some point

I don't know much about it to fix it, probably should just return nullptr when it can't get/find it, right?

Dregu: Well then using any of the get_entity functions very early probably crashes too, but there's no reason to use them before entities can exist so I don't see a problem. Avoiding crashes would be nicer though.

Dregu: Returning nullptr and making the functions using it aware of this to return nil or {} or whatever would be great.

Mr-Auto: `get_entity` is actually using local, it's hard to see because there is `State::find` and `State::find_local`
very confusing but `find` uses `ptr()` which just returns `ptr_local()`, not sure why @Malacath-92 left the `ptr()` instead of always using either `ptr_local()` or `ptr_main()`

Dregu: Yeah almost everything is using local, ptr was pointed to that a long time ago. I reckon it was just easier than chaging all the instances to _local. Just added _main in a the few places local doesn't make sense.

Dregu: `state`, `players` and `prng` are the only things always on main, which probably causes issues (online) when those things are used from a callback that's running on a different thread.

PR #303, opened 2023-02-16
Fix it

Malacath-92:
Don't have time to figure out what the params mean, but checking them via
```lua
    ent:set_post_damage(function(...)
        prinspect("damage: ", ...)
    end)
end, SPAWN_TYPE.ANY, MASK.ANY)
```
seems like they all are sensible.


Dregu:
This doesn't fix the issue of poison passing through monsters hooked by on_damage instead of poisoning them, in release build anyway.

Auto already fixed the funny params in #301, but reverted the optimization hack you didn't want.

Malacath-92:
Oh yes, I derped hard and while verifying that the crucial part was actually the issue I didn't revert back to it.

Dregu:
Derp, well that makes more sense. What does it return though? More unknown flags?

Dregu:
Ok yeah so you can play with the flags to ignore fire damage, so you only take 10 damage from bombs, still die in lava, for example:

```lua
P:set_pre_damage(function(a,b,c,d,e,f,g,h,i)
  if test_mask(d, DAMAGE_TYPE.FIRE) then
    d = clr_mask(d, DAMAGE_TYPE.FIRE)
    if d > 0 then
      return a:regular_damage(b,c,d,e,f,g,h,i)
    else
      return 0
    end
  end
end)
```

We're really collecting broken Movable::damage functions though.

Malacath-92:
From Ghidra it looked like it returned something based on the entities flags. Maybe the flags that are changing or the resulting flags or the overlap of damage_flags and its own flags. Smth smth

Dregu:
Well now we're thinking it's bool "yep, I got hit or something"

PR #307, opened 2023-03-01
Safe set_setting

Dregu:
- Adds a subset of visual settings SAFE_SETTING
- Exposes set_setting that can be used during the level for temporary changes
- Settings are reset to real values and can't be set while in the options screen*

*I think the cfg files are only ever saved if settings are edited directly in the options menu, so the files should be protected from mod edits


Dregu:
This shouldn't be very controversial, will merge for next nightly.

PR #314, opened 2023-06-20
Expose more fields and update win_pushid

Cosine256:
This PR exposes some miscellaneous fields, updates `win_pushid` to also accept strings, fixes an EmmyLua script bug, and updates some `docs/game_data` files that had fallen behind by several commits.


src/game_api/script/usertypes/state_lua.cpp
```diff
@@ -167,6 +168,8 @@ void register_usertypes(sol::state& lua)
     /// Used in StateMemory
     lua.new_usertype<Items>(
         "Items",
+        "leader",
```
Dregu, line 171: So we usually try to keep the player index related things in the range 1..4 in lua, and this one is probably 0..3. 1..4 indexes could be used directly in get_player etc...

I say try, cause we may have slipped just as many 0-based indexes in the api as there are 1-based ones.

Anyway, this is what we do to fix some of them: https://github.com/spelunky-fyi/overlunky/blob/daa21b8a96c5fd95e278cc8dc336edc4927f873a/src/game_api/script/usertypes/entities_chars_lua.cpp#L48-L60

I dunno, maybe documenting the range would be enough.

Cosine256: Yeah, the range for that is 0..3. It makes sense to adjust the range for Lua if we've been doing that elsewhere, so I'll make that change. I have `leader` as an unsigned value, but I see `player_slot` is signed. Is there any convention for what integer type to use in this situation?

Dregu: The player slot in the inventory can be -1 for AI, but that's not possible for the leader so it's fine as unsigned.

Cosine256: Do we care about overflow for this leader field, or can I just let the integer arithmetic do its thing if the user sets it to something outside the range 1..4 in Lua?

Mr-Auto: sol should take care of this, if you expose 8bit field, it should only allow to modify those 8bit, as the memory is not actually "exposed" to the end user

Cosine256: That's what it does if I don't handle it in any way. The unsigned byte wraps from 0 to 255. I meant whether we wanted to do anything about the edge cases instead of letting it wrap. I was just planning to add 1 in the getter and -1 in the setter.

Cosine256: I pushed a commit with the change I intended to make.

PR #315, opened 2023-07-10
Add ImGui widgets

Cosine256:
Exposes some ImGui widgets and features to the Lua API:

* Tooltips
* Color editor
* _WIP_

### DIFF
--- a/examples/window.lua
+++ b/examples/window.lua
@@ -10,6 +10,8 @@ local inputdrag = 9
 local inputcombo = 1
 local comboopts = { 'one', 'two', 'three' }
 local inputcheck = false
+local inputcolorrgb = Color:green()
+local inputcolorrgba = Color:new(0.0, 0.0, 1.0, 0.5) -- 50% opacity blue
 
 local inputtooltip = 'This input\'s value is used as its tooltip.'
 
@@ -69,6 +71,8 @@ set_callback(function(draw_ctx)
             inputdrag = draw_ctx:win_drag_float('Select another number##drag', inputdrag, 1, 10)
             inputcombo = draw_ctx:win_combo('Combo thing', inputcombo, table.concat(comboopts, '\0')..'\0\0')
             inputcheck = draw_ctx:win_check('Check this out', inputcheck)
+            inputcolorrgb = draw_ctx:win_color_editor('Color (RGB)', inputcolorrgb, false)
+            inputcolorrgba = draw_ctx:win_color_editor('Color (RGBA)', inputcolorrgba, true)
 
             draw_ctx:win_separator_text('Section Two')
             -- pseudo table layout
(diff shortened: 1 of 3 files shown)


Dregu:
I recently struggled a bit making good ui with my window api, so I'm just leaving a note here for someone to implement everything in [this thread](https://stackoverflow.com/questions/72840675/how-to-create-a-vertical-group-of-buttons-in-imgui/72840945) too. Not a request, just a reminder. You might enjoy #337 for the TAS2L btw.

Cosine256:
It looks like all that's needed for that feature is a `BeginGroup`/`EndGroup` call, so I think I can add that without any trouble. If there was anything else from ImGui that you or anybody else wanted, I can add it to my list if it isn't terribly complicated.

My design philosophy for these changes has been to make the GuiDrawContext API closely imitate the ImGui API. Is there any reason not to do it this way? I feel there is no reason to create our own facade for the underlying library. I'm still adding "simple" overloads for people who just want to cobble together a basic GUI using only the OL docs, while power users can look at ImGui docs if they want to fully understand the overloads with ImGui flags and other arguments.

Also, I saw the request for exposing a safe file directory and I definitely appreciate it. It'll be nice to not require unsafe mode for the TAS Tool.

Dregu:
I didn't mean just some way to make a vertical group, I meant every function mentioned in the thread. The big buttons, way to draw more things over a cursorpos saved earlier etc.

Anyway, my idea is also to closely imitate the imgui C api, while not directly exposing any Begins and Ends, cause forgetting one of those would lead to a disaster. But you're already wrapping these to callbacks and know what I'm talking about. I guess some imgui features were just deemed unnecessarily complicated earlier, but hopefully those can still be fixed with overloads for advanced arguments.

Cosine256:
Why are borders disabled on windows even if the OL setting is enabled? It's difficult to read overlapping windows since they seamlessly blend into each other.

https://github.com/spelunky-fyi/overlunky/blob/f44e576dad16d148c8aface07b55821ca13dc334/src/injected/ui.cpp#L654

Cosine256:
I exposed some of the ImGui size and position calls, including the ones mentioned in that Stack Overflow thread. A problem I'm seeing with them is that they're in either ImGui screen or window coordinates (pixels). You can't easily use them with the `GuiDrawContext:draw_*` calls if you wanted to do something like draw a circle within the space created by a `win_dummy`, because those are all using OL screen coordinates. Should those calls convert to OL screen coordinates? Should there be some kind of way for the user to do that conversion themselves? I'm also considering renaming the "cursor" ones, since ImGui's name is confusing and refers to the ImGui output position instead of a mouse cursor.

Dregu:
> Why are borders disabled on windows even if the OL setting is enabled?

It's probably just a leftover from before the option was even added, and maybe I was going for no borders just on the OL *menus*, but disabled the completely wrong one and forgot about it. You can change that line to follow the UI option. Also we should add a safe way to use PushStyleVar and other Push-able styling functions in scripts i.e. automatically popping any stray pushes at the end of the gui context.

> Should those calls convert to OL screen coordinates?

Yes, but there are probably always cases were pixel format would be useful, like using the cursor pos in sameline, which uses pixels. Even worse that the imgui coordinates are desktop-based, not window-based like our api.

> Should there be some kind of way for the user to do that conversion themselves?

Yes, per previous answer. Adding a pixel variant for every function using screen coordinates surely isn't the way, so exposing converters is the least we can do.

> I'm also considering renaming the "cursor"

I agree, but don't know what it would be. Maybe output_position like you said, or even input_position, cause it is used to position input fields usually.

Dregu:
I realized after the PushStyleVar comment that Begin stacks can also be counted in the gui ctx, so wrapping all of them in callbacks isn't the only way to make sure they are Ended properly at the end of the window or ctx. I can look into this later.

Also all gui callbacks need to be forwarded the ctx as first parameter (`handle_function<...>(backend, callback, this)`) in case the callback function is not inline and can't use the upvalue from the GUIFRAME callback. My fault, I forgot it in `win_section` too, but at least it's in `window`.

Dregu:
Here's the idea from the previous comment with win_disabled to optionally use callbacks or just BeginDisabled that will be fixed automatically if forgotten by modder. https://github.com/Cosine256/overlunky/compare/add-imgui-widgets...spelunky-fyi:overlunky:GuiStackFix

PR #323, opened 2023-09-02
Map/set stuff also skip in layer

Mr-Auto:


Mr-Auto:
Wondering if we can use this stuff to improve functions like `get_entities_overlapping_hitbox` and `get_entities_at`

Mr-Auto:
Wondering if there is a need for listing all the owned items, like `get_owned_items`, to get all the uid's from the `state->room_owners->owned_items`.

Maybe that would be too confusing with the `owner` and `last_owner` i `Movable` ?

Dregu:
> Wondering if there is a need for listing all the owned items, like `get_owned_items`, to get all the uid's from the `state->room_owners->owned_items`.
> 
> Maybe that would be too confusing with the `owner` and `last_owner` i `Movable` ?

Sounds useful, if only for some science.

src/game_api/script/usertypes/state_lua.cpp
```diff
@@ -554,6 +555,15 @@ void register_usertypes(sol::state& lua)
...
     logicdiceshop_type["balance"] = &LogicDiceShop::balance;
 
+    /// Used in StateMemory
+    lua.new_usertype<RoomOwnersInfo>("RoomOwnersInfo", "owned_items", &RoomOwnersInfo::owned_items, "owned_rooms", &RoomOwnersInfo::owned_rooms);
+
+    /// Used in RoomOwnersInfo
+    lua.new_usertype<ItemOwnerDetails>("ItemOwnerDetails", "owner_type", &ItemOwnerDetails::owner_type, "owner_uid", &ItemOwnerDetails::owner_uid);
+
+    /// Used in RoomOwnersInfo
+    lua.new_usertype<RoomwOwnerDetails>("RoomwOwnerDetails", "layer", &RoomwOwnerDetails::layer, "room_index", &RoomwOwnerDetails::room_index, "owner_uid", &RoomwOwnerDetails::owner_uid);
```
Dregu, line 565: Oh looks like Roomw has a typo everywhere and I didn't notice a single one when I checked this earlier

PR #328, opened 2023-09-08
fix PRNG::get_local not getting the local prng

estebanfer:
the prng wasn't set to the current local one due to it being static, returning basically the main prng instead


Dregu:
Oh oops. Maybe this invalidated all my online rando tests.

PR #330, opened 2023-09-12
New community tilecodes

estebanfer:
I don't really know what would be the conditions to accept new community tilecodes, but I'll leave this PR here :cat2:
I have already a branch of ML locally with the new tilecodes on the editor, if that's needed, should I open a PR on ML too?
Also, I don't know if I should include the `parachute_attach` tilecode? it simply deploys a parachte on the entity below


Dregu:
I don't consider myself the authority for accepting community tilecodes, but these seem simple enough... The parachute maybe more like a gimmick.

PR #332, opened 2023-09-24
Movable vtable hooks and fix virtual indexes

Dregu:
I didn't really test any of this yet besides process_input, it's fun to control everything.

```lua
set_post_entity_spawn(function(e)
  e:set_pre_process_input(function(e)
    e.buttons = P.buttons
    e.movex = P.movex
    e.movey = P.movey
    return true
  end)
end, SPAWN_TYPE.ANY, MASK.MONSTER | MASK.MOUNT | MASK.ACTIVEFLOOR)
```


src/game_api/movable.hpp
```diff
@@ -167,7 +167,7 @@ class Movable : public Entity
     virtual void drop(Entity* entity_to_drop) = 0;               // 69, also used when throwing
 
     /// Adds or subtracts the specified amount of money to the movable's (player's) inventory. Shows the calculation animation in the HUD.
-    virtual void add_money(uint32_t money) = 0;                              // 70
+    virtual void add_money(uint32_t money, ENT_TYPE money_type) = 0;         // 70
```
Mr-Auto, line 170: btw. this is only used for collecting the gems and gold. All the other stuff, like selling the idol or buying something in shop is done by a different function.

Dregu: an unknown function or..?

Mr-Auto: Well, there is no clear one function to do this. Like normally in the game money can be only subtracted when you buy something from the shop. 
The best solution would probably be to find/make function that displays the difference on the hud and then edit the amount of money in the `state.money_shop_total` as it's the only one that can go negative

Mr-Auto: btw. it is known that the total current money is?:
loop all 4 players:
[ inventory.money + inventory.money_collected_total ]
+state.money_shop_total 

The game doesn't even check if the other players are playing or not, and does move their inventory.money to inventory.money_collected_total (probably since in multiplayer they can disconnect)

Mr-Auto: Even funnier thing: when you sell idol in the shop, it does not call the add_money function but the idol is added to the inventory as collected money

Dregu: Yeah that calculation is even documented somewhere in the docs

Mr-Auto: ok found the hud thing, it's two static values. One for the amount of money, the other for the time it's screen

Dregu: you mean the rendering thing we already have in the api?

Mr-Auto: maybe? didn't really pay attention it the API thing since it was just in the rendering callback

Mr-Auto: Well, still need a pattern for it anyway if we want a proper standalone add_money function

Dregu: I think this is basically what you had to do, too complicated

```lua
set_callback(function(ctx, hud)
  if add ~= 0 then
    hud.data.money_total = hud.data.money_total + add
    hud.data.money_counter = add
    hud.data.money.counter = add
    hud.data.money.timer = 48
    players[1].inventory.money = players[1].inventory.money + add
    add = 0
  end
end, ON.RENDER_PRE_HUD)

function add_money(n)
  add = n
end
```

Mr-Auto: So i think it would be good to write some explanation for the `add_money` function, or even change the name of the virtual, so the `add_money` will only be the broken* one

Like `collect_treasure(int32_t worth, ENT_TYPE treasure_type` maybe?

also it's funny that the comment mentions subtracting but the parameter is unsigned 😆

Dregu: Hmm yeah technically fixing the add_money function to always subtract from the treasure count is a breaking change, even though it crashes if used incorrectly, but maybe we'll just leave it like it was. I like collect_treasure more too.

Dregu:
Well I tested all the movable hooks I left in. The ones I removed are either broken, wrong, useless or unknown.

PR #333, opened 2023-09-25
Work on logic stuff

Mr-Auto:
The `// OK` comment is just for me, will be removing it at the end


src/game_api/state_structs.hpp
```diff
@@ -388,99 +389,106 @@ struct ArenaState
...
+    SPEEDRUN,
+    GHOST,
+    GHOST_TOAST,
+    TUN_AGGRO,
+    DICESHOP,
+    PRE_CHALLENGE,
+    MOON_CHALLENGE,
+    STAR_CHALLENGE,
+    SUN_CHALLENGE,
+    MAGMAN_SPAWN,
```
Dregu, line 405: MAGMAMAN

docs/parse_source.py
```diff
@@ -891,7 +892,8 @@ def run_parse():
                 if not var:
                     continue
                 var = var.split(",")
-                vars.append({"name": var[0], "type": var[1]})
+                if(len(var) > 1):
+                    vars.append({"name": var[0], "type": var[1]})
```
Mr-Auto, line 896: Apparently the situation is: `var[0] = '//'` and no `var[1]`

PR #335, opened 2023-09-27
Add http GET api

Dregu:
```lua
http_get_async("https://mossranking.com/api/getwr.php?cat=201", function(res, err)
    if res then
        local wr = json.decode(res)
        prinspect(wr)
    elseif err then
        prinspect(err)
    end
end)
```

`{ "S̼̼ͪh͚͚ͯa͔͔ͩd̶̵ow", "81250", "1:21.250", "2023-08-05" }`


Dregu:
Time flies, I wrote this last year it seems. It was just missing the safe_cb wrapper, which it now has. Probably wasn't anything else wrong with it.

PR #339, opened 2023-10-01
Docs validator

Mr-Auto:


src/game_api/script/usertypes/vanilla_render_lua.cpp
```diff
@@ -790,8 +790,6 @@ void register_usertypes(sol::state& lua)
         &TextRenderingInfo::height,
         "special_texture_id",
         &TextRenderingInfo::special_texture_id,
-        "font",
-        &TextRenderingInfo::font,
```
Mr-Auto, line 794: rip if anyone was using this, thou I doubt you could even do anything with it

Mr-Auto:
list as of right now:
`'SpawnInfo', 'set', 'HttpRequest', 'UdpServer', 'customtheme', 'size_t'`

- `SpawnInfo` - no idea, little documentation, no examples
- `set` - we have maps, sets, vectors, tables, arrays - would be great to decide on like one name, or because of the standard objects come with some extra functions, maybe would be good to not rename them and just write a documentation?
- `HttpRequest` and `UdpServer` - no idea
- `customtheme` - should just ignore? it's in the `force_custom_subtheme` and `force_custom_theme`, as they are written in a "non standard" way compared to other stuff
- `size_t` - if i remember correctly, sol2 doesn't handle unsigned 64bit well, also this is only used for `get_address` and `get_rva` which i don't think should be in the API doc, as almost no one except people who work on this project/reverse engineer the game use this

Dregu:
We can stop renaming the stl containers and link to https://sol2.readthedocs.io/en/latest/containers.html in the first section for reference.

I think the idea of returning the socket types was to transfer ownership or something, but I don't really know how it all works out.

Why keep the debug functions a secret though?

Mr-Auto:
> We can stop renaming the stl containers and link to https://sol2.readthedocs.io/en/latest/containers.html in the first section for reference.

Don't think this has any value to someone that just wants to write a mod and have no knowledge about C++, standard library etc.

> Why keep the debug functions a secret though?

Hard to keep something secret in open source project 😄 . It's just that normal people don't have use for those. And for example: if i ever needed to get address for pattern, i would never think to look in the API doc for something like this

docs/src/includes/_types.md
```diff
@@ -418,8 +418,8 @@ Stores static common data for an [ENT_TYPE](#ENT_TYPE). You can also clone entit
 
 Type | Name | Description
 ---- | ---- | -----------
-[EntityDB](#EntityDB) | [new(EntityDB other)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=EntityDB) | 
-[EntityDB](#EntityDB) | [new(ENT_TYPE)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=EntityDB) | 
+[EntityDB](#EntityDB) | [new(EntityDB& other)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=EntityDB) | 
+[EntityDB](#EntityDB) | [new(ENT_TYPE other)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=EntityDB) |
```
Mr-Auto, line 422: did i accidentally fix something?
or break something?

Dregu: Looks like a fix to me...

Dregu:
> Don't think this has any value to someone that just wants to write a mod and have no knowledge about C++, standard library etc.

Just write a doc section about the container types yourself then if you think that's not good enough. It's would probably be enough for most people if we just say that all of these containers can be used like lua tables, but maybe add a matrix of available operators/methods too.

> Hard to keep something secret in open source project 😄 . It's just that normal people don't have use for those. And for example: if i ever needed to get address for pattern, i would never think to look in the API doc for something like this

I don't know where else you would look for something like this, but I'd rather remove the deprecated functions before this, and I don't want to do that either. If it's in the api, it should be in the api doc. It's just more confusing if people find these functions and we intentionally left them undocumented.

docs/src/includes/_types.md
```diff
@@ -583,32 +583,33 @@ prinspect(string.format("%x"), color:get_ucolor()) -- 800000ff
...
+  [Color](#Color) | [olive()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=olive) | 
+  [Color](#Color) | [lime()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=lime) | 
+  [Color](#Color) | [green()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=green) | 
+  [Color](#Color) | [aqua()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=aqua) | 
+  [Color](#Color) | [teal()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=teal) | 
+  [Color](#Color) | [blue()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=blue) | 
+  [Color](#Color) | [navy()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=navy) | 
+  [Color](#Color) | [fuchsia()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=fuchsia) | 
+  [Color](#Color) | [purple()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=purple) | 
 tuple&lt;int, int, int, int&gt; | [get_rgba()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_rgba) | Returns RGBA colors in 0..255 range
```
Mr-Auto, line 609: autodoc why you like this?

estebanfer:
> * `set` - we have maps, sets, vectors, tables, arrays - would be great to decide on like one name, or because of the standard objects come with some extra functions, maybe would be good to not rename them and just write a documentation?

arrays and tables can be different tho, `luals` has some advanced types related to that https://luals.github.io/wiki/annotations/#documenting-types

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -1326,7 +1326,7 @@ void register_usertypes(sol::state& lua)
     lua.create_named_table("DYNAMIC_TEXTURE", "INVISIBLE", DYNAMIC_TEXTURE::INVISIBLE, "BACKGROUND", DYNAMIC_TEXTURE::BACKGROUND, "FLOOR", DYNAMIC_TEXTURE::FLOOR, "DOOR", DYNAMIC_TEXTURE::DOOR, "DOOR_LAYER", DYNAMIC_TEXTURE::DOOR_LAYER, "BACKGROUND_DECORATION", DYNAMIC_TEXTURE::BACKGROUND_DECORATION, "KALI_STATUE", DYNAMIC_TEXTURE::KALI_STATUE, "COFFIN", DYNAMIC_TEXTURE::COFFIN);
 
     /// Force a theme in PRE_LOAD_LEVEL_FILES, POST_ROOM_GENERATION or PRE_LEVEL_GENERATION to change different aspects of the levelgen. You can pass a CustomTheme, ThemeInfo or THEME.
-    // lua["force_custom_theme"] = [](* customtheme)
+    // lua["force_custom_theme"] = [](mixed* customtheme)
```
estebanfer, line 1329: you could use `CustomTheme|ThemeInfo|THEME` instead of `mixed`

docs/src/includes/_types.md
```diff
@@ -2833,7 +2851,7 @@ nil | [spawn_transition()](https://github.com/spelunky-fyi/overlunky/search?l=Lu
 nil | [post_transition()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=post_transition) | 
 nil | [spawn_players()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=spawn_players) | 
 nil | [spawn_effects()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=spawn_effects) | 
-string | [get_level_file()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_level_file) | 
+char | [get_level_file()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_level_file) |
```
Mr-Auto, line 2854: ...

Dregu:
I am using the address functions through string.format("%x") always though, so returning hex strings automatically would just make it easier.

docs/src/includes/_globals.md
```diff
@@ -627,7 +627,7 @@ Get uids of matching entities overlapping with the given hitbox. Set `entity_typ
 
 > Search script examples for [get_entity](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_entity)
 
-#### [Entity](#Entity) get_entity(int uid)
+#### [Entity](#Entity)  get_entity(int uid)
```
Mr-Auto, line 630: another weird thing to the collection

docs/src/includes/_types.md
```diff
@@ -840,10 +841,10 @@ float | [tileh](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=tileh)
...
-class [Entity](#Entity) | [get_entity()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_entity) | 
+[Entity](#Entity) | [get_entity()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_entity) | 
 bool | [set_normal_map_texture(TEXTURE texture_id)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_normal_map_texture) | Sets second_texture to the texture specified, then sets third_texture to SHINE_0 and texture_num to 3. You still have to change shader to 30 to render with normal map (same as COG normal maps)
-optional&lt;[TEXTURE](#TEXTURE)&gt; | [get_second_texture](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_second_texture) | 
-optional&lt;[TEXTURE](#TEXTURE)&gt; | [get_third_texture](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_third_texture) | 
+std::optional&lt;[TEXTURE](#TEXTURE)&gt; | [get_second_texture](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_second_texture) | 
+std::optional&lt;[TEXTURE](#TEXTURE)&gt; | [get_third_texture](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_third_texture) |
```
Mr-Auto, line 847: huh

docs/src/includes/_types.md
```diff
@@ -1669,8 +1678,8 @@ Used in [ScreenCharacterSelect](#ScreenCharacterSelect), [ScreenTitle](#ScreenTi
 Type | Name | Description
 ---- | ---- | -----------
 [ParticleDB](#ParticleDB) | [particle_type](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=particle_type) | 
-int | [particle_count](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=particle_count) | 
-int | [particle_count_back_layer](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=particle_count_back_layer) | 
+uint32_t | [particle_count](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=particle_count) | 
+uint32_t | [particle_count_back_layer](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=particle_count_back_layer) |
```
Mr-Auto, line 1682: huh x2

docs/src/includes/_globals.md
```diff
@@ -522,7 +522,7 @@ Remove item by uid from entity
 
 > Search script examples for [filter_entities](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=filter_entities)
 
-#### array&lt;int&gt; filter_entities(array<int> entities, function predicate)
+#### vector&lt;int&gt; filter_entities(vector<int> entities, function predicate)
```
Mr-Auto, line 525: Think i will just leave this

Mr-Auto:
We can use * in the comments again, instead of this funny snowflake ✱

Mr-Auto:
['HttpRequest', 'UdpServer'] left

docs/generate_emmylua.py
```diff
@@ -14,16 +14,16 @@
     "int32_t": "integer",
     "int64_t": "integer",
     "size_t": "integer",
+    "in_port_t": "integer",
     "ImU32": "integer",
     "vector": "Array",
     "array": "Array",
```
Mr-Auto, line 20: why those `Array` is the only one that starts from big letter? it that even important?

docs/game_data/spel2.lua
```diff
@@ -6320,9 +6335,9 @@ function CustomTheme:new() end
 Vec2 = nil
 ---@return Vec2
 function Vec2:new() end
----@param vec2 Vec2
+---@param 2 Vec
 ---@return Vec2
-function Vec2:new(vec2) end
+function Vec2:new(2) end
```
Mr-Auto, line 6340: Broken for some reason

docs/src/includes/_types.md
```diff
@@ -614,6 +614,7 @@ Type | Name | Description
 ---- | ---- | -----------
 [Color](#Color) | [new()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=Color) | Create a new color - defaults to black
 [Color](#Color) | [new(Color)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=Color) | 
+[Color](#Color) | [new(Color color)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=Color) |
```
Mr-Auto, line 617: some oddity

docs/game_data/spel2.lua
Dregu: Well yeah spel2.lua as a whole just seems worse now with these glitches, can't really merge that.

docs/generate_emmylua.py
```diff
@@ -92,7 +92,7 @@ def replace_all(text):
 ps.run_parse()
 
 
-reGetParam = re.compile(r"(?!const)(\b[^ ]+) *([^,]+),?")
+reGetParam = re.compile(r"(?!const)(\b[^ ]+) *([^,]+),?") # this pattern have some issues
```
Dregu, line 95: We're using regex to interpret a whole language, issues is all we have :D

Mr-Auto:
> I think the idea of returning the socket types was to transfer ownership or something, but I don't really know how it all works out.
> 

So should i make them return nil or do actually ownership transfer and ignore the type in the validator?

Mr-Auto:
right now there is no ownership, so the pointer is just dangling, if i do ownership transfer it will mean that using just `udp_listen("url", foo)` without saving the return will prompt sol to just delete the object

so it's kind of breaking change, but probably the right solution

Dregu:
The idea was to transfer the UdpServer to lua so it could have methods like close()... I just didn't add the type cause there were no methods yet and lua usertypes are only required by this validator. You can add an empty type or even implement :close() for it, definitely not return nil.

HttpRequest is a similar story, although it has a pretty short lifespan and doesn't really matter what it returns, as the callback handles success and error. As long as it doesn't turn to garbage before the request even finishes.

Dregu:
And it's fine if garbage collection closes the udpserver when the lua handle is gone, that's how it's already documented but it just doesn't work like that.

src/game_api/script/usertypes/level_lua.cpp
```diff
@@ -1807,5 +1807,16 @@ void register_usertypes(sol::state& lua)
     // DEFAULT
     // FLOOR | SAFE | EMPTY
     */
+
+    lua.new_usertype<SpawnInfo>(
+        "SpawnInfo",
+        "room_temp",
```
Dregu, line 1813: Probably shouldn't expose as "room_temp", it's ambiguous and called "room_template" everywhere else.

PR #340, opened 2023-10-05
Random UI things

Dregu:
- Random UI fixes
- Expose `update_state()` to run the game logic really fast, maybe to skip forward in a TAS or double the game speed
- Add `set_frametime()` to change the target engine fps, capped by gpu/monitor/vsync/whatever still

```lua
-- skip hundun rise in a few seconds, simulating everything in between but not wasting time to render it
while state.time_level < 5000 do
  update_state()
end

-- run game at double speed, but keep timer at normal speed
set_callback(function()
  if not no_repeat then
    no_repeat = true
    update_state()
    state.time_level = state.time_level - 1
    state.time_total = state.time_total - 1
    no_repeat = nil
  end
end, ON.POST_UPDATE)

-- run at 50% speed by changing frametime
set_frametime(1/30)

-- don't cap fps when window is unfocused
set_frametime_unfocused(0)
```

### DIFF
--- a/src/game_api/custom_types.cpp
+++ b/src/game_api/custom_types.cpp
@@ -551,7 +551,8 @@ std::span<const ENT_TYPE> get_custom_entity_types(CUSTOM_TYPE type)
             "ENT_TYPE_ITEM_DMCRATE",
             "ENT_TYPE_ITEM_PRESENT",
             "ENT_TYPE_ITEM_GHIST_PRESENT",
-            "ENT_TYPE_ITEM_ALIVE_EMBEDDED_ON_ICE");
+            "ENT_TYPE_ITEM_ALIVE_EMBEDDED_ON_ICE",
+            "ENT_TYPE_ITEM_POT");
     case CUSTOM_TYPE::CONVEYORBELT:
         return make_custom_entity_type_list<CUSTOM_TYPE::CONVEYORBELT>(
             "ENT_TYPE_FLOOR_CONVEYORBELT_LEFT",
--- a/src/game_api/flags.hpp
+++ b/src/game_api/flags.hpp
@@ -754,5 +754,5 @@ const char* pause_types[]{
     "8: ?",
     "16: ?",
     "32: Ankh (smooth camera, janky audio)",
-    "Freeze on PRE_UPDATE",
+    "Freeze on PRE_UPDATE", // this is not a real state.pause flag, it's only used by ui.cpp for magic
 };
--- a/src/game_api/flags.hpp
+++ b/src/game_api/flags.hpp
@@ -754,4 +754,5 @@ const char* pause_types[]{
     "8: ?",
     "16: ?",
     "32: Ankh (smooth camera, janky audio)",
+    "Freeze on PRE_UPDATE",
 };
--- a/src/game_api/movable.hpp
+++ b/src/game_api/movable.hpp
@@ -163,7 +163,7 @@ class Movable : public Entity
     virtual void on_picked_up_by(Entity* entity_picking_up) = 0;
     virtual void drop(Entity* entity_to_drop) = 0; // also used when throwing
 
-    /// Adds or subtracts the specified amount of money to the movable's (player's) inventory. Shows the calculation animation in the HUD. Important: Updates the collected_money_count in the inventory, but doesn't add any treasure to the following array, which will lead to crashes during the transition, if you don't deal with it. (e.g. set `collected_money_count = collected_money_count - 1` afterwards)
+    /// Adds or subtracts the specified amount of money to the movable's (player's) inventory. Shows the calculation animation in the HUD.
     virtual void add_money(uint32_t money) = 0;
 
     virtual void apply_movement() = 0;              // disable this function and things can't move, some spin in place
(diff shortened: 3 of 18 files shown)


src/game_api/movable.hpp
```diff
@@ -163,7 +163,7 @@ class Movable : public Entity
     virtual void on_picked_up_by(Entity* entity_picking_up) = 0;
     virtual void drop(Entity* entity_to_drop) = 0; // also used when throwing
 
-    /// Adds or subtracts the specified amount of money to the movable's (player's) inventory. Shows the calculation animation in the HUD.
+    /// Adds or subtracts the specified amount of money to the movable's (player's) inventory. Shows the calculation animation in the HUD. Important: Updates the collected_money_count in the inventory, but doesn't add any treasure to the following array, which will lead to crashes during the transition, if you don't deal with it. (e.g. set `collected_money_count = collected_money_count - 1` afterwards)
     virtual void add_money(uint32_t money) = 0;
```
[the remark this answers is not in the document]
Mr-Auto: should be `virtual void add_money(uint32_t money, ENT_TYPE money_type) = 0;`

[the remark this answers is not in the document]
Dregu: Well that's not a surprise, the whole vtable is still a big mess.

src/game_api/custom_types.cpp
```diff
@@ -551,7 +551,8 @@ std::span<const ENT_TYPE> get_custom_entity_types(CUSTOM_TYPE type)
             "ENT_TYPE_ITEM_DMCRATE",
             "ENT_TYPE_ITEM_PRESENT",
             "ENT_TYPE_ITEM_GHIST_PRESENT",
-            "ENT_TYPE_ITEM_ALIVE_EMBEDDED_ON_ICE");
+            "ENT_TYPE_ITEM_ALIVE_EMBEDDED_ON_ICE",
+            "ENT_TYPE_ITEM_POT");
```
Mr-Auto, line 555: This is technically is not correct, can you maybe create a new type for this, could then also include coffin

src/game_api/flags.hpp
```diff
@@ -754,4 +754,5 @@ const char* pause_types[]{
     "8: ?",
     "16: ?",
     "32: Ankh (smooth camera, janky audio)",
+    "Freeze on PRE_UPDATE",
```
Mr-Auto, line 757: missing the actual flag number?

src/game_api/script/lua_vm.cpp
```diff
@@ -2014,31 +2014,46 @@ end
...
 
     /// Allows you to disable the control over the door for Hundun and Tiamat
     /// This will also prevent game crashing when there is no exit door when they are in level
     lua["set_boss_door_control_enabled"] = set_boss_door_control_enabled;
 
+    /// Run state update manually, i.e. simulate one logic frame. Use in e.g. POST_UPDATE, but be mindful of infinite loops, this will cause another POST_UPDATE. Can even be called thousands of times to simulate minutes of gameplay in a few seconds.
+    lua["update_state"] = update_state;
+
+    /// Set engine target frametime (1/framerate, default 1/60). Always capped by your GPU max FPS / VSync. To run the engine faster than rendered FPS, try update_state. Set to 0 to go as fast as possible. Call without arguments to reset.
+    lua["set_frametime"] = set_frametime;
```
Mr-Auto, line 2046: I guess this invalidates the `CONST.ENGINE_FPS`

src/game_api/flags.hpp
Dregu: Not a real flag, just a way to pause.

src/game_api/custom_types.cpp
Dregu: Neither is present as far as I know. I still think "can contain arbitrary item" would be a good definition for a container, and coffin doesn't belong there, cause non-chars crash the game.

src/game_api/flags.hpp
Mr-Auto: So only the UI thing? maybe add a comment about this, since we link the `flags.hpp` in the API doc as well

src/game_api/custom_types.cpp
Dregu: And they all have `ENT_TYPE inside`, which is the important part, they can all be used blindly exactly the same way...

Dregu: So is present actually a container, or should we remove the presents too, or just add Pot, cause they're really all the same thing anyway?

Mr-Auto: The idea of custom types was, you can take the class name and use it as `ENT_TYPE` to get all entities as listed in `entity_hierarchy`. And also add extra ones that have similar properties etc.

present is there because this was made before I added the `Purchasable` class

So ideally this should only have: `ITEM_CRATE, ITEM_DMCRATE, ITEM_ALIVE_EMBEDDED_ON_ICE`

Dregu: Because of the exactly similar properties, and because removing present would be a breaking change anyway, I'm keeping pot.

PR #341, opened 2023-10-07
Add destroy_grid_entity function

estebanfer:
Adds the `destroy_grid_entity` function to be able to remove a grid entity from the grid at the same time it gets destroyed, so the neighbor entities can be fixed right after destroying it, and a way to correctly destroy grid entities before the game has created their decorations, to have the correct decorations
I was thinking I could add some optional parameter to disable destroying the items of the grid entity, so if the user wants to do something different with the items, it could be done by the user itself (?)


Dregu:
I'd also suggest a `destroy_grid(x, y, l)` so you don't have to figure out what's there when playing with coordinates.

src/game_api/script/lua_vm.cpp
```diff
@@ -966,6 +966,9 @@ end
     lua["move_entity"] = move_entity_abs;
     /// Teleport grid entity, the destination should be whole number, this ensures that the collisions will work properly
     lua["move_grid_entity"] = move_grid_entity;
+    /// Destroy the grid entity, and its item entities, removing them from the grid without dropping particles or gold.
+    /// Will also destroy monsters or items that are standing on a linked activefloor or chain, though excludes MASK.PLAYER to prevent crashes
+    lua["destroy_grid_entity"] = destroy_grid_entity;
```
Mr-Auto, line 971: activefloors are not grid entities

src/game_api/layer.cpp
```diff
@@ -215,3 +215,35 @@ void Layer::move_grid_entity(Entity* ent, uint32_t x, uint32_t y, Layer* dest_la
...
+void Layer::destroy_grid_entity(Entity* ent)
+{
+    if (ent)
+    {
+        auto items = ent->items.entities();
+        for (auto ptr = items.cend(); ptr != items.cbegin();)
+        {
+            ptr--;
+            Entity* item_ent = *ptr;
+            if (item_ent->type->search_flags & ~1) // if not player
```
Mr-Auto, line 228: this feels very weirdly specific

src/game_api/layer.cpp
```diff
@@ -215,3 +215,35 @@ void Layer::move_grid_entity(Entity* ent, uint32_t x, uint32_t y, Layer* dest_la
...
+
+        const auto pos = ent->position();
+        const uint32_t current_grid_x = static_cast<uint32_t>(std::round(pos.first));
+        const uint32_t current_grid_y = static_cast<uint32_t>(std::round(pos.second));
+        if (current_grid_x < g_level_max_x && current_grid_y < g_level_max_y)
+        {
+            if (grid_entities[current_grid_y][current_grid_x] == ent)
+            {
+                grid_entities[current_grid_y][current_grid_x] = nullptr;
+                update_liquid_collision_at(pos.first, pos.second, false);
```
Mr-Auto, line 242: isn't this done automatic by the destroy function?

src/game_api/script/lua_vm.cpp
Dregu: Activefloors can be overlayed on grid floors, like chain blocks...

src/game_api/layer.cpp
Dregu: Destroying the player would just crash the game. Although ent->is_player() is a better check for this, like used in UI::safe_destroy

Mr-Auto: I didn't mean this as: why check player, but like: why only check player?

src/game_api/layer.cpp
Dregu: No, destroy doesn't update liquid collision, kill does.

src/game_api/layer.cpp
Dregu: As a safeguard, cause it's the only thing that crashes the game. Also I think destroying the items is usually the right thing to do, like with chainbugs on a chain on the chain ceiling you're distroying, but crashing is definitely never what you want.

Dregu: This is pretty theoretical, cause it's a level gen function and would ideally be used before monsters spawn or even chains are grown, but we can still prepare for that.

estebanfer:
> I'd also suggest a `destroy_grid(x, y, l)` so you don't have to figure out what's there when playing with coordinates.

So, a different function with that name right? Or an overload?

Dregu:
> > I'd also suggest a `destroy_grid(x, y, l)` so you don't have to figure out what's there when playing with coordinates.
> 
> So, a different function with that name right? Or an overload?

Well the one you already made clearly expects an entity(uid), but I wouldn't mind if this new function has an overload that treats a single argument as uid and 3 as coordinates...

estebanfer:
It feels a bit weird for me since most overloads have almost the same kind of parameters or just extra ones, though `create_illumination` already does take either coordinates or uid.
What do you think would be better?

Dregu:
>  What do you think would be better?

Well I think of it this way: What is the harm if we add the overload? Will someone accidentally forget the rest of the arguments and something terrible happens when the uid was supposed to be an x-coordinate, or is it just a convenient shorthand with no risk of confusing anyone...

estebanfer:
well, doesn't make sense passing uid, y position and layer, or passing just x position, hopefully people won't be that dumb :cat2:

estebanfer:
oh and this reminds me that overloads generated in `spel2.lua` are kinda broken... I'll have to fix that

Dregu:
> oh and this reminds me that overloads generated in `spel2.lua` are kinda broken... I'll have to fix that

Ah yes the new safe io and os stuff is also missing, I was supposed to make something for it, but it's probably easiest to just manually add `io.open_data = io.open` or something.

Dregu:
Still suggesting the overloaded function to be called `destroy_grid`, as `destroy_grid_entity` clearly expects an entity...

estebanfer:
ohh... true

PR #342, opened 2023-10-07
Some random things that i can't think of single name for

Mr-Auto:
- add_custom_type
- get_entities_by_draw_depth
- kill_recursive
- destroy_recursive


docs/src/includes/_types.md
```diff
@@ -3776,6 +3776,10 @@ nil | [set_invisible(bool value)](https://github.com/spelunky-fyi/overlunky/sear
 array&lt;int&gt; | [get_items()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_items) | 
 bool | [is_in_liquid()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=is_in_liquid) | Returns true if entity is in water/lava
 bool | [is_cursed()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=is_cursed) | 
+nil | [kill_recursive(bool destroy_corpse, Entity responsible, optional<int> mask, const array<ENT_TYPE> ent_types, RECURSIVE_MODE rec_mode)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=kill_recursive) | Kill entity along with all entities attached to it. Be aware that for example killing push block with this function will also kill anything on top of it, any items, players, monsters etc.<br/>To a that, you can inclusively or exclusively limit certain [MASK](#MASK) and [ENT_TYPE](#ENT_TYPE). Note: the function will first check mask, if the entity doesn't match, it will look in the provided [ENT_TYPE](#ENT_TYPE)'s<br/>destroy_corpse and responsible are the standard parameters for the kill funciton
```
Mr-Auto, line 3779: > To a that

"a-void"
🤣

Mr-Auto: This should be fixed in #339

Dregu: Maybe we should just avoid that word so we can avoid fixing the terrible python mess

Mr-Auto:
Tested a little with all kinds of setups. Also the `destroy_recursive` very nicely gets rid of the bosses (Olmec, Tiamat, Kingu)

Dregu:
> Tested a little with all kinds of setups. Also the `destroy_recursive` very nicely gets rid of the bosses (Olmec, Tiamat, Kingu)

But probably not megajelly... It's weird af and the tail is not attached to anything. ui_util has a funny method to destroy it though, might want to add that here.

Dregu:
UI::safe_destroy also has this philosophy that you can select something really stupid like the bosses left arm, and it will first find the parent and then recursively destroy the parts. That's not really the point of this function, but if we add the api safe_destroy, this is the way.

Mr-Auto:
> > Tested a little with all kinds of setups. Also the `destroy_recursive` very nicely gets rid of the bosses (Olmec, Tiamat, Kingu)
> 
> But probably not megajelly... It's weird af and the tail is not attached to anything. ui_util has a funny method to destroy it though, might want to add that here.

That one person who coded the CO is crazy, orbs being tracked by `uid + (n)`, tail that is clearly attached to an entity is actually not and the entities just have the variables chain xd

Mr-Auto:
Also one thing that I'm not sure of:
in `INCLUSIVE` mode, you have to have the mask or type of the main entity in the parameters to actually kill/destroy it, and similar with `EXCLUSIVE` if you have the entity type or mask in the parameters it won't kill/destroy it either

But it feels like it should always kill/destroy the main entity, should I change that?

Mr-Auto:
> UI::safe_destroy also has this philosophy that you can select something really stupid like the bosses left arm, and it will first find the parent and then recursively destroy the parts. That's not really the point of this function, but if we add the api safe_destroy, this is the way.

I kind of stayed away from the `safe_destroy` as it's actually super complicated, with some situations that aches your brain with "what if the user actually wants it this way". Thou i always wanted to do `safe_spawn`, but never actually took the time to do it

Mr-Auto:
> > Tested a little with all kinds of setups. Also the `destroy_recursive` very nicely gets rid of the bosses (Olmec, Tiamat, Kingu)
> 
> But probably not megajelly... It's weird af and the tail is not attached to anything. ui_util has a funny method to destroy it though, might want to add that here.

So i made the code, but then started thinking, maybe i should write a patch for this instead? then we could have a jellyfish with shorter tail, or even more with some more code

Dregu:
> But it feels like it should always kill/destroy the main entity, should I change that?

I can't think of a reason right now, but it does sound useful to be able to only destroy some matching parts but not the main entity.

Mr-Auto:
> > But it feels like it should always kill/destroy the main entity, should I change that?
> 
> I can't think of a reason right now, but it does sound useful to be able to only destroy some matching parts but not the main entity.

that's not what the function will do. If the entity in excluded or not included it will not be killed/destroyed along with the items attached to it.

So for example: if you kill push block with item rock included, it will not kill the rock that the player standing on the pushblock is holding

Mr-Auto:
Could also just expose the `get_hud` function or even make a global variable like `state` and `players`

Dregu:
> Could also just expose the `get_hud` function or even make a global variable like `state` and `players`

Exposing the function is fine.

This add_money script should be just part of the api, adding the shop total automatically. The hud function is fine, but nobody really needs to jump through those hoops.

Mr-Auto:
> > Could also just expose the `get_hud` function or even make a global variable like `state` and `players`
> 
> Exposing the function is fine.
> 
> This add_money script should be just part of the api, adding the shop total automatically. The hud function is fine, but nobody really needs to jump through those hoops.

I didn't want to decide if it should add money to the `inventory.money` or `state.money_shop_total` since someone can use both for something, making separate counters for each player on how much money they collected for example. So they want the money to actually be added to the specific player inventory and not the money_shop

Dregu:
> So they want the money to actually be added to the specific player inventory and not the money_shop

Ok, how about `add_money(amount)` goes to the total and `add_money(amount, playerid)` goes to the inventory.

Mr-Auto:
> > So they want the money to actually be added to the specific player inventory and not the money_shop
> 
> Ok, how about `add_money(amount)` goes to the total and `add_money(amount, playerid)` goes to the inventory.

`playerslot` instead of `playerid`?

Dregu:
> `playerslot` instead of `playerid`?

Yeah that, I don't even know what playerid is.

Mr-Auto:
ok, so it's all numbers so i can't make overload with the optional `display_time`

Dregu:
> ok, so it's all numbers so i can't make overload with the optional `display_time`

Eh I don't think that's super useful and it can still be changed with the hud functions.

Dregu:
So at least doors don't crash, but we still need

```lua
set_post_entity_spawn(function(e)
  e:set_pre_on_collision2(function(_, d)
    return not test_flag(d.flags, ENT_FLAG.ENABLE_BUTTON_PROMPT)
  end)
end, SPAWN_TYPE.ANY, MASK.ANY, ENT_TYPE.PET)
```

to make the dogs act normally, i.e. just keep walking past closed doors.

Mr-Auto:
> So at least doors don't crash, but we still need
>  ...
> to make the dogs act normally, i.e. just keep walking past closed doors.

Probably easier this way than to just patch another game code

Mr-Auto:
Why do i get build error about something i didn't do?

Dregu:
Life is unfair

Mr-Auto:
+1k, think I'm not with this PR

Dregu:
I've been trying to keep up with it, but I think I'll need to read some of those lines again tomorrow...

docs/game_data/spel2.lua
```diff
@@ -1235,6 +1235,39 @@ function set_frametime_unfocused(frametime) end
 ---Get engine target frametime when game is unfocused (1/framerate, default 1/33).
 ---@return double?
 function get_frametime_unfocused() end
+---Adds new custom type (group of ENT_TYPE) that can be later used in functions like get_entities_by or set_(pre/post)_entity_spawn
+---@param types ENT_TYPE[]
+---@return ENT_TYPE
+function add_custom_type(types) end
```
Dregu, line 1241: So if this was to be used with the "clone entitydb and spoof id scenario", it would probably be nice if it could be called without arguments (should probably also add the usual variadic ENT_TYPEs) to simply reserve an unused id for a new type, that can then be nicely found with the get functions. Something like

```lua
reserved_id = add_custom_type()
custom_rock = EntityDB:new(ENT_TYPE.ITEM_ROCK)
custom_rock.id = reserved_id
some_rock.type = custom_rock
get_entities_by_type(reserved_id) -- some_rock
```

src/game_api/level_api.cpp
```diff
@@ -420,20 +429,14 @@ std::array g_community_tile_codes{
...
-            layer->spawn_apep(x, y, !is_room_flipped(x, y));
-        }},
+    CommunityTileCode{"apep_left", "ENT_TYPE_MONS_APEP_HEAD", []([[maybe_unused]] const CommunityTileCode& self, float x, float y, Layer* layer)
+                      {
+                          layer->spawn_apep(x, y, is_room_flipped(x, y));
+                      }},
+    CommunityTileCode{"apep_right", "ENT_TYPE_MONS_APEP_HEAD", []([[maybe_unused]] const CommunityTileCode& self, float x, float y, Layer* layer)
+                      {
+                          layer->spawn_apep(x, y, !is_room_flipped(x, y));
+                      }},
```
Mr-Auto, line 439: no idea why visual studio decided to format those differently

PR #344, opened 2023-10-08
fix spel2.lua function overloads

estebanfer:
Add `---@meta` to spel2.lua, wich indicates the language server that it is a definition file, making the multiple function declarations to work correctly
https://luals.github.io/wiki/annotations/#meta


estebanfer:
oh, `@meta`  in the commit seems to point to an user in github, lol

PR #345, opened 2023-10-08
I guess we're doing UI again

Dregu:
- UI: Add level gen flags to game props
- UI: Add warp buttons for other types of game screens
- UI: Add option to disable death screen popping up
- UI: Add button/hotkey to respawn dead players
- UI: Add monster/trap/room/generic procedural chances to game props
- API: Add `create_level/layer()`, initialize layers back, create an empty level after you've destroyed it
- API: Add `destroy_level/layer()`, wipes all entities and unloads the layer(s)
- API: Add `set_level_logic_enabled()`, disable all usual death screens from popping up, including when playing with 0 players
- API: Add `Movable:clear_cutscene()` to stop intro, transition etc cutscenes
- API: Add helpers to convert between INPUTS and (x, y, BUTTON)
- API: Add `ON.PRE/POST_LAYER/LEVEL_CREATION`, called before/after any screen creates layers
- API: Add `ON.PRE/POST_LAYER/LEVEL_DESTRUCTION`, called before/after any screen unloads a layer or both layers
- API: Make `ON.PRE_LEVEL_GENERATION` cancelable
- API: Make `ON.*_LEVEL_GENERATION` run in transitions, since it runs in almost every other screen anyway

```lua
-- convert the current level to an empty void with just the player
destroy_level()
create_level()
state.current_theme:spawn_players()

-- same sort of thing with the level gen hook
set_callback(function()
  state.illumination = create_illumination(Color:white(), 20000, 172, 252)
  state.current_theme:spawn_players()
  return true
end, ON.PRE_LEVEL_GENERATION)
```


Dregu:
I hope the cutscene allocation works like I think, even if the CustomCutscene alloc is bigger than the vanilla, and vanilla does the destroying. At least using game_allocator stopped it from crashing on dtor.

Mr-Auto:
> I hope the cutscene allocation works like I think, even if the CustomCutscene alloc is bigger than the vanilla, and vanilla does the destroying. At least using game_allocator stopped it from crashing on dtor.

There are like 3 cutscenes in the game, we can probably map all of them if you want. The question is, do we really need this? it's nothing special, just a function that is called every frame, and it disables the entity logic (probably just skips calling some virtuals like process input)

Dregu:
Well I'm sure we don't need it, there are ton of places to hook for similar results. My question is really is there anything wrong with it, now that it's already done? Or can you show me a simpler way to replicate the functionality with existing methods?

src/game_api/spawn_api.cpp
```diff
@@ -662,18 +662,33 @@ void init_spawn_hooks()
...
-    using spawn_player_fun = void(Items*, uint8_t ps, float pos_x, float pos_y);
+    auto old_x = state->level_gen->spawn_x;
+    auto old_y = state->level_gen->spawn_y;
+    state->level_gen->spawn_x = x.value_or(old_x);
+    state->level_gen->spawn_y = y.value_or(old_y);
+    auto uid = (int32_t)state->next_entity_uid;
+    using spawn_player_fun = void(Items*, uint8_t ps);
     static auto spawn_player = (spawn_player_fun*)get_address("spawn_player");
-    spawn_player(get_state_ptr()->items, player_slot - 1, x, y);
+    spawn_player(get_state_ptr()->items, player_slot - 1);
```
Dregu, line 688: I'm pretty sure it didn't even have those params and all players just spawned at the level_gen->spawn coordinates anyway.

Mr-Auto: Yes, it only takes 2 parameters, is mostly called from the virtual populate_players in theme
For the layer hack, it doesn't cover the HH spawn, player ghost and much more (powerups maybe?)

Dregu: What do those things have to do with spawning a player in the back layer manually?

Mr-Auto: The function does not spawn player only, it spawn all the stuff from the inventory, so all the power ups, all the hired hand that were assigned to the player the previews levels etc.

Dregu: Oh right, I didn't really read it through. Not very applicable to the scenario I was thinking about, but it would probably be a problem in let's say some interconnected world where you can enter and leave the level from back layer with some items too, not just debugging in OL

Mr-Auto: Function is in the API, if it was only for debugging/UI I would not be concerned
It's probably possible to make the layer stuff work, but it needs more code, probably the easiest would be to look thru it and change all the `0x1300` patterns, + the HH probably needs the `patch_and_redirect` as there is not enough space to set the layer param for that function

Dregu: ok hear me out

Mr-Auto: That's so stupid, i love it xd
thou you probably should not use ` write_mem_recoverable` as it's saves the old memory once, so if the pointer changes it will recover to the wrong address

Mr-Auto: Also no need to hardcode `0x1300` offset in this case, you can just change the layer pointers around

Dregu: Yeah I don't know why I wrote it like that

src/game_api/spawn_api.hpp
```diff
@@ -49,7 +49,7 @@ void pop_spawn_type_flags(SPAWN_TYPE flags);
 
 void init_spawn_hooks();
 
-void spawn_player(int8_t player_slot, float x, float y);
+int32_t spawn_player(int8_t player_slot, std::optional<float> x = std::nullopt, std::optional<float> y = std::nullopt, std::optional<LAYER> layer = std::nullopt);
```
Mr-Auto, line 52: autodoc will probably not like defaults, and kind of no reason to have them as sol doesn't care and you wrote separate UI version

src/game_api/search.cpp
```diff
@@ -2034,6 +2034,50 @@ std::unordered_map<std::string_view, AddressRule> g_address_rules{
...
+        "dead_players"sv,
+        // TODO
+        PatternCommandBuffer{}
+            .from_exe_base(0x22c061d0),
+    },
+    {
+        "spawn_transition"sv,
+        // TODO
+        PatternCommandBuffer{}
+            .from_exe_base(0x22afe5c0),
```
Mr-Auto, line 2061: That's theme virtual 21

Mr-Auto: Same for the rest of transitions, just different theme virtual, there are 5:
standard, jungle and volcana (have some extra code before jumping to the standard function),
Co duat and hundun use the same one, there is also basecamp one, not sure for what for, and city of gold (probably deciding which one to call, standard or to duat)

Dregu: I know they're all theme virtual 21, and I specifically didn't hook the vtable for all themes (like in LevelGenSystem::hook_theme) cause that didn't work. I don't know why, but get_entities() was empty in the callbacks made this way, even though it worked fine when using state.current_theme:set_post_spawn_transition()

Dregu: And obviously this callback would be useless if you can't find the entities just spawned...

Mr-Auto: > I know they're all theme virtual 21, and I specifically didn't hook the vtable for all themes (like in LevelGenSystem::hook_theme) cause that didn't work. I don't know why, but get_entities() was empty in the callbacks made this way, even though it worked fine when using state.current_theme:set_post_spawn_transition()

That's super weird, you sure you didn't made off by one error or something? or maybe state local is busted when initiating level?
If not, maybe we can use `get_virtual_function_address` or something like that here, instead of harcoded offset

src/game_api/spawn_api.cpp
```diff
@@ -662,18 +662,38 @@ void init_spawn_hooks()
...
     push_spawn_type_flags(SPAWN_TYPE_SCRIPT);
     OnScopeExit pop{[]
                     { pop_spawn_type_flags(SPAWN_TYPE_SCRIPT); }};
 
-    using spawn_player_fun = void(Items*, uint8_t ps, float pos_x, float pos_y);
+    auto old_x = state->level_gen->spawn_x;
+    auto old_y = state->level_gen->spawn_y;
+    state->level_gen->spawn_x = x.value_or(old_x);
+    state->level_gen->spawn_y = y.value_or(old_y);
+    auto uid = (int32_t)state->next_entity_uid;
```
Mr-Auto, line 685: HH are spawned first, if you really need the uid returned you can just get it from `state.items.players`

Dregu: It just seemed weird that it's the only spawn function that doesn't return uid and after learning all the things it really spawns, still sounds logical to return the player uid.

src/game_api/search.cpp
Dregu: It was pretty late then, but after testing again I am still pretty confident I'm hooking the right function, cause it still doesn't work if I call current_theme:spawn_transition() directly. Sounds like get_virtual_function_address would simplify this though.

src/game_api/search.cpp
```diff
@@ -2034,6 +2034,50 @@ std::unordered_map<std::string_view, AddressRule> g_address_rules{
...
+        "init_layer"sv,
+        // TODO
+        PatternCommandBuffer{}
+            .from_exe_base(0x228b58f0),
+    },
+    {
+        "dead_players"sv,
+        // TODO
+        PatternCommandBuffer{}
+            .from_exe_base(0x22c061d0),
```
Mr-Auto, line 2055: And this is screen level `v1` virtual

Mr-Auto: Also death screen is `v4`
i should probably look more into those

Mr-Auto: btw. screen in state, not in gamemaganer, we should really make some clear distinctions hmm

Dregu: I was just going to make patterns, but how is using hardcoded VTABLE_OFFSETs any better than using hardcoded RVAs though?

Mr-Auto: vtable offsets i can update with one click in the plugin

Dregu: So it's still worse than pattern, assuming the pattern didn't break on update, and better than pattern if the pattern did break on update...

Mr-Auto: and both are better than harcoded offsets :)

i don't mind patterns, thou using a vtable offset makes it also clear where to look for it if it breaks

Mr-Auto: Also the vtable offsets don't break every update

src/game_api/script/usertypes/vtables_lua.cpp
```diff
@@ -114,6 +114,9 @@ void register_usertypes(sol::state& lua)
         {
             Entity* ent = get_entity_ptr(uid);
             entity_vtable.unhook(ent, callback_id);
+            movable_vtable.unhook(ent, callback_id);
+            floor_vtable.unhook(ent, callback_id);
+            door_vtable.unhook(ent, callback_id);
```
Dregu, line 119: I don't understand this template jungle, but clear_callback didn't work and this fixed it.

Dregu:
Ok I'm done with this pr, it's getting pretty hefty.

src/game_api/script/lua_vm.cpp
```diff
@@ -2096,6 +2100,53 @@ end
...
+    lua["destroy_layer"] = destroy_layer;
+
+    /// Initializes an empty front and back layer that don't currently exist. Does nothing(?) if layers already exist.
+    lua["create_level"] = create_level;
+
+    /// Initializes an empty layer that doesn't currently exist.
+    lua["create_layer"] = create_layer;
+
+    /// Setting to false disables the death screen from popping up for any usual reason, can still load manually
+    lua["set_death_enabled"] = set_death_enabled;
```
Dregu, line 2116: Ok so this also disables some small things like shop zoom, but do we care? I made it just for overlunking, but it's nice to have debugging tools in the api too.

Mr-Auto: It's also just nice to not need a player in a level at all

Dregu: Yes for cutscenes or weird game modes in mods too I suppose. I'll change the name to something more descriptive, any ideas?

Mr-Auto: Dunno, also going back to the patterns, we could just get the address from vtable of state.screen_level, and for the theme stuff from level_gen 
since they never gets destroyed and stuff

Dregu: I don't really want to redo the hooks for a fourth time, but you can suggest your fix in pr format and this works for now.

PR #352, opened 2023-11-02
Bucket

Dregu:
- API: Add a `Bucket` thing in shared memory for interoperability between the modding tools for Cosine256/spelunky-2-tas-wizard#1
  - API: Add `set_start_level_paused` to start levels paused
  - API: Add `zoom_reset` to be able to reset default shop zoom
  - API: Add `get_global_frame` for a frame counter that doesn't pause ever
  - API: Expose hookable menu input vars in GameProps
  - API: Fix `get_type` crashing with custom type > 915
  - API: Fix `set_camera_position` for moving the camera faster or during pause

Why do I have to write something between two lists github?

  - UI: Add renderer flags to game tab
  - UI: Add absolute player camera hack to camera tab
  - UI: Add bomb fuse to hovered entity tooltip
  - UI: Add text filter for finding scripts
  - UI: Add button to change script directory
  - UI: Fix crust item spawning to not trigger when spawning with an offset
  - UI: Fix spawned over eggsac angle
  - UI: Fix spawning items with invalid type id


docs/parse_source.py
```diff
@@ -11,6 +11,56 @@
 if not os.path.exists(".db"):
     os.makedirs(".db")
 
+replace_table = {
+    # standard basic types
+    "uint8_t": "int",
+    "uint16_t": "int",
+    "uint32_t": "int",
+    "uint64_t": "int",
```
Mr-Auto, line 19: Best to probably move it to separate file completely, so no one will be tempted to just use it here directly

src/game_api/entity_db.cpp
```diff
@@ -97,7 +97,7 @@ EntityDB* get_type(uint32_t id)
     // Special case: map_ptr might be 0 if it's not initialized.
     // This only occurs in list_entities; for others, do not check the pointer
     // to see if this assumption works.
-    if (!entity_factory_ptr)
+    if (!entity_factory_ptr || id > 0x393)
```
Mr-Auto, line 100: We always tented to use `to_id` instead of hard coded type id, but as we don't see any updates in the future it's probably fine for hardcoded values now

docs/parse_source.py
Dregu: Oh maybe, it just seemed silly that you have to edit it in three places when 2 need to be the same.

src/game_api/entity_db.cpp
Dregu: Yeah the actual array size seemed hardcoded anyway and even if a new entity was added, we woudn't know what it was and couldn't use to_id to get the last id.

Mr-Auto: well, not often they add new entity at the end (as they did in last big update xd)
btw. this can still crash as there are gaps in the id's, so it's not even size, it's just last id

src/game_api/bucket.hpp
```diff
@@ -0,0 +1,47 @@
...
+    /// Currently selected uid in the entity picker or -1 if nothing is selected.
+    int32_t selected_uid;
+    /// Set currently selected uid in the entity picker or -1 to clear selection.
+    std::optional<int32_t> set_selected_uid;
+    /// Currently hovered entity uid or -1 if nothing is hovered.
+    int32_t hovered_uid;
+    /// Currently selected uids in the entity finder.
+    std::vector<uint32_t> selected_uids;
+    /// Set currently selected uids in the entity finder.
+    std::optional<std::vector<uint32_t>> set_selected_uids;
```
Mr-Auto, line 35: You could probably tread this as just shared memory and add `bool patches_applied{0};` and replace the watermark code in state.cpp with checking and setting this, so we only need this one place to put the pointer

Dregu: Yeah that was the idea too that we can move any checks like that to the bucket. I was actually just going to use the original watermark location first, but I think there was no room for a pointer there.

PR #354, opened 2023-11-05
Raw input

Dregu:
Tried to figure out more of the input tables.

- API: Add `get_raw_input` -> RawInput* to view all pressed keyboard keys and controller buttons without imgui
- API: `ON.PRE/POST_PROCESS_INPUT` to disable or capture all input
- UI: Add player input selector to game props


src/game_api/game_manager.hpp
```diff
@@ -70,6 +70,43 @@ struct BackgroundMusic
...
+    std::array<ControllerInput, 12> controller;
+};
+
+struct SomeInput
+{
+    bool enabled;
+    uint8_t padding1[3];
+    int8_t input_index;
+    uint8_t padding2[3];
+    uint32_t buttons;
```
Mr-Auto, line 107: The struct is 
4x bool
int8 input_index
//probably padding
int32 buttons
int8 unknown

Mr-Auto: the first bool is disabled not enabled
second one is like enabled or something (setting false disables it)
dunno about third
fourth is "connected" - if set to false you get popup about controller disconnected

Mr-Auto: ok, third one is like "can_use_menu"

Dregu: Oh yeah I figured there must be more of them, but didn't actually even test what the first one does. The controllers kept getting reserved, not being able to switch to again, even though not set in any input_index, and some of these is probably why

Mr-Auto: The last byte is used as first parameter to `XInputSetState`, at least for the XInput controllers

Mr-Auto: also the buttons are just copy over from `GameProps.buttons`, it's just x4 that, then x4 dupe (for each player), then the last one is for the menu

Mr-Auto: The difference between this and state.inputs buttons i guess is for example: those direct inputs, while the state have the run button held all the time if you have the setting always run enabled

Mr-Auto: Also buttons and buttos_dupe are used to determinate held status, since buttons_dupe are just buttons 1 frame later

Mr-Auto: also also, all of them are really 16bits, not 32, thou not sure if it matters

Mr-Auto: and the menu has to be 16bit since it's buttons_one_frame_later and buttons, and yes, in that order, so it's not nice and easy xd

Mr-Auto: didn't know how to name all
image

Dregu: Yeah I already have that enum in the api for menu input

Mr-Auto: btw. what `modal_open` suppose to be?
i think is just the player slot used for the icon in pause menu

Dregu: > the first bool is disabled not enabled
> second one is like enabled or something (setting false disables it)
> dunno about third
> fourth is "connected" - if set to false you get popup about controller disconnected

these didn't make any sense after all, everything you said seems backwards

Dregu: > btw. what `modal_open` suppose to be?
> i think is just the player slot used for the icon in pause menu

Yeah I didn't know what it actually does, and that sounds logical, but I just used it to detect if the menu or the journal is open, cause that's when that icon is set too.

Mr-Auto: the fo

> > the first bool is disabled not enabled
> > second one is like enabled or something (setting false disables it)
> > dunno about third
> > fourth is "connected" - if set to false you get popup about controller disconnected
> 
> these didn't make any sense after all, everything you said seems backwards

the fourth i made mistake it's more like "lost_connection" the rest should be correct, so by default you have:
```
false
true
true
false
```

Dregu: Except for keyboard those are not default

Mr-Auto: i guess we're wrong about the first two, and keyboard doesn't seam to care about the menu one
or the struct are different for keyboard vs controller

Dregu: Whatever, not really even smart to expose that struct to the api...

Mr-Auto: also the struct is much larger, seen accessed qword at offset 0x280

Dregu:
This is still in a bit of a weird state, but I think I'll merge it cause I may have merged even dumber stuff earlier and this fixes some of that.

src/game_api/script/usertypes/game_manager_lua.cpp
```diff
@@ -95,20 +97,67 @@ void register_usertypes(sol::state& lua)
...
-        &GameProps::buttons_extra,
+        [](GameProps& gp) -> uint32_t
+        {
+            return gp.buttons[1];
+        },
+        "buttons_previous",
+        [](GameProps& gp) -> uint32_t
+        {
+            return gp.buttons_previous[1];
+        },
```
Mr-Auto, line 117: shouldn't this be `[0]` ? also probably best to `NoDoc` it

src/game_api/script/usertypes/game_manager_lua.cpp
```diff
@@ -95,20 +97,67 @@ void register_usertypes(sol::state& lua)
...
         &GameProps::buttons_menu_previous,
-        "buttons_menu",
+        "input",
+        &GameProps::buttons,
+        "input_previous",
+        &GameProps::buttons_previous,
+        "input_menu",
         &GameProps::buttons_menu,
+        "input_menu_previous",
+        &GameProps::buttons_menu_previous,
```
Mr-Auto, line 129: why `buttons_menu` and `input_menu` ?

Dregu: I may have copypasted too many things. Yeah I just wanted to deprecate stupid singular buttons and couldn't use the same name and didn't want to call the array buttonses like gollum.

PR #356, opened 2023-11-07
More TAS hooks

Dregu:
UI:
- Add pause option that can pause and frame advance during menus, fades etc.
- Add pause option to auto-pause on level start
- Add pause option to skip freezing updates on loads and fades when player interaction is impossible
- Add set seed and adventure options to game tab
- Add cheat option to skip all fades
- Add camera option to set speed multiplier (for a bit faster or near instant tas presentation camera)
- Add camera option to apply smooth camera during update pauses
- Fix quick start not initializing all the correct things
- Fix respawn players option to work in the death screen
- Add automatic version check next to version overlay
- Add Seed menu to Game tab, to get, set or restart current adventure seed
- Add ability to figure out and restart current adventure seed even if OL is loaded during the run

API:
- Add `ON.PRE/POST_GAME_LOOP`, like PRE_UPDATE but can pause menu and other UI updates
- Add `update_camera_position` to apply the automatic tweening and level bounds checks manually
- Add `set_speedhack` to control the speedhack
- Add methods to get/set the seed entered in ScreenSeedInput
- Add optional argument for `get_adventure_seed` to return the seed at the beginning of current run


src/game_api/memory.cpp
```diff
@@ -103,6 +103,7 @@ LPVOID alloc_mem_rel32(size_t addr, size_t size)
 }
 
 std::unordered_map<std::string, std::vector<RecoverableMemory>> original_memory;
+std::unordered_set<std::string> edited_memory;
```
Mr-Auto, line 106: Feel like better solution would be to change `std::vector<RecoverableMemory>` into a struct with vector and bool

src/game_api/memory.cpp
```diff
@@ -147,9 +149,15 @@ void recover_mem(std::string name, size_t addr)
         for (auto& it : original_memory[name])
             if (!addr || addr == it.address)
                 write_mem_prot(it.address, std::string_view{it.old_data, it.size}, it.prot_used);
+        edited_memory.erase(name);
```
Mr-Auto, line 152: should be only erased when `addr == 0` (whole name, not just single addr)

src/game_api/rpc.cpp
```diff
@@ -1868,3 +1868,139 @@ void set_start_level_paused(bool enable)
...
+}
+
+void init_adventure()
+{
+    // TODO: I didn't check exactly what this does, but it fixes issues with character select being broken after quick start
+    static size_t offset = 0;
+    if (offset == 0)
+    {
+        offset = get_address("init_adventure");
+    }
```
Mr-Auto, line 1984: Why not just  `static size_t offset = get_address("init_adventure");` ?

src/game_api/script/lua_vm.cpp
```diff
@@ -680,12 +680,14 @@ end
     /// Show a message that looks like a level feeling.
     lua["toast"] = [](std::u16string message)
     {
+        State::get().ptr()->toast = 0;
```
Mr-Auto, line 683: Won't this cause memory leak?
also it changes the behavior, for example the `say` function have some sort of buffer, so if you call it couple of times the second message will show up after the first one disappears

src/game_api/rpc.cpp
Dregu: I dunno, we have this old crap laying around all over the place and I'm not really thinking when I copy paste it and change the pattern

src/game_api/memory.cpp
Dregu: You're

src/game_api/memory.cpp
Dregu: right

src/game_api/script/lua_vm.cpp
Dregu: Oh I assumed you'd always want to display a new message with these and the one put on the screen by the game was just an annoyance. At least that's probably the case with toast. The cancel_functions should still be fixed so they can be used on the same frame, but `delete` didn't fix the memory leak either. Reverted anyway.

Dregu: You know

```lua
toast("some boring vanilla feeling set by the vanilla level gen")
cancel_toast() -- should cancel immeditely, i.e. set the toast to null so toast() can be used again immediately
toast("custom feeling should always override vanilla feeling, at least after cancel_toast, without waiting for next update")
```

Dregu:
That's enough random (tas) stuff for this week.

PR #364, opened 2023-11-26
Stuff

Mr-Auto:
Let me know if i should rename the `special_sound` to `lise_sound` or something, as it only apply to that one character


src/game_api/level_api.hpp
```diff
@@ -142,7 +142,7 @@ struct LevelGenData
             uint32_t machine_rewardroom_chance;
             uint32_t max_liquid_particles;
             uint32_t flagged_liquid_rooms;
-            uint32_t unknown_config;
+            uint32_t unknown_config; //
```
Mr-Auto, line 145: forgot to write the comment, it's padding, not a real setting

PR #365, opened 2023-11-27
Pause fixes

Cosine256:
This PR fixes a few issues loosely related to the OL pause system:

- Improves StateMemory fade variable names and documentation.
- Fixes OL freeze pauses to consistently stop on the last fade-in frame of a screen.
- Updates OL freeze pauses to also stop on the first fade-out frame of a screen.
- Fixes a bug where bucket `options` took an extra render cycle to update after processing `set_options`.

### DIFF
--- a/src/game_api/state.hpp
+++ b/src/game_api/state.hpp
@@ -64,12 +64,27 @@ struct StateMemory
     uint32_t loading;
     /// The global level illumination, very big and bright.
     Illumination* illumination;
-    /// Current fade-to-black amount (0.0 = all visible; 1.0 = all black). Manipulated by the loading routine when loading > 0.
-    float fadevalue;
-    /// Amount of frames the fadeout should last when loading
-    uint32_t fadeout;
-    /// Amount of frames the fadein should last when loading
-    uint32_t fadein;
+    union
+    {
+        /// Current fade-to-black amount (0.0 = all visible; 1.0 = all black). Manipulated by the loading routine when loading > 0.
+        float fade_value;
+        /// NoDoc
+        float fadevalue;
+    };
+    union
+    {
+        /// Remaining frames for fade-in/fade-out when loading. Counts down to 0.
+        uint32_t fade_timer;
+        /// NoDoc
+        uint32_t fadeout;
+    };
+    union
+    {
+        /// Total frames for fade-in/fade-out when loading.
+        uint32_t fade_length;
+        /// NoDoc
+        uint32_t fadein;
+    };
     /// if state.loading is 1, this timer counts down to 0 while the screen is black (used after Ouroboros, in credits etc.)
     uint32_t loading_black_screen_timer;
     /// Is 1 when you in a game, is set to 0 or 1 in main menu, can't be trusted there, normally in a level is 1 unless you go to the options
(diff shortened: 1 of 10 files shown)


src/game_api/state.hpp
```diff
@@ -64,12 +64,27 @@ struct StateMemory
...
+        /// NoDoc
+        uint32_t fadeout;
+    };
+    union
+    {
+        /// Total frames for fade-in/fade-out when loading.
+        uint32_t fade_length;
+        /// NoDoc
+        uint32_t fadein;
+    };
```
Mr-Auto, line 87: I would just rename them, then in `state_lua.cpp` you can do the NoDoc

Dregu: Yep you can remove the unions, change all the internal references to the new names and add the deprecated lua bindings with NoDoc. There are probably examples somewhere.

Dregu:
I've been thinking about and trying to fix some edge cases that are still present, like

- unpausing during loading should not touch the `state.pause` flags,
- fade pause is required to work with level start pauses,
- level start pause gets enabled for all screens, but only some actually react to that pause type,
- UI only detects level start pauses in level screen,
- other situations where UI pause state gets desynced from the script...

I've come to the conclusion, that

- `set_start_level_paused` is stupid and very confusing if not in level screen,
- touching `state.pause` at all makes everything more complicated,
- touching `state.pause` is obsolete after adding the camera update function.

To fix the situation, I think UI should

- just support the two freeze pause types properly in C side callbacks,
- add a custom implementation for different auto-pause conditions,
- let scripts control the new pause system via bucket like before,
- let the game handle state.pause like it wants and leave it alone.

I can probably look at those things this week. Meanwhile this PR does fix some things, but requires TASW or the user in general to use very specific OL settings to not get stuck paused somewhere...

Cosine256:
For a user wanting to do pausing and frame advancing, is there anything `state.pause` manipulation accomplishes that can't be done with the freeze pauses? Deprecating and abandoning the `state.pause` system would be so much simpler. Scripts can still change `state.pause` themselves if they want to utilize the quirks of those pause types for stuff like custom cutscenes, but I don't see how it's useful for frame advancing when we have much better freeze pauses now.

The only thing I can think of is that changing `state.pause` in a script applies instantly to the next game update instead of needing a render like `set_options` in the bucket does. Is it possible for modifying the bucket to apply instantly? Could the bucket directly expose `paused`, `g_pause_type`, and all of the other useful OL variables that are in `ui.cpp`?

The more work that goes into this, the less it makes sense for me to make TASW not dependent on OL. When should an OL feature become a script API feature instead, and exist whether or not OL is attached? Things like the entity GUIs and the hitbox viewer feel like they belong in OL, but this pause system is looking more like a script API feature, like `modify_sparktraps` and `zoom`, or that weird `pause` function that I doubt anybody uses. Would moving the freeze pause system into the script API cause problems? Couldn't Playlunky and Overlunky share the same state for these features via the bucket?

Dregu:
> Is there anything `state.pause` manipulation accomplishes that can't be done with the freeze pauses? 

Not anymore, since we can update the camera manually and don't need to rely on a pause flag that doesn't break the camera.

> Would moving the freeze pause system into the script API cause problems? Couldn't Playlunky and Overlunky share the same state for these features via the bucket?

Yeah sounds like the best option, then the UI can just control it using the same api as the scripts, shared pause state saved in the bucket. The set-ui-options part of the bucket is becoming obsolete really fast, but that's just a good thing.

> Is it possible for modifying the bucket to apply instantly?

When the api can write directly to the shared pause options, it's probably as instant as it can get. Then it's just a question of which callback you are setting it in. Could probably even toggle the pre-update pause inside the pre-update callback of the same frame if the api pause is checked after script callbacks.

Dregu:
I'll merge this for the other fixes and start looking at better pause api today

PR #367, opened 2024-01-05
Fix the hooks in level_api.hpp to avoid the code divergence

Malacath-92:


Dregu:
This got reverted cause it apparently broke some mods, namely randomizer was skipping all SPAWN_TYPE.SYSTEMIC callbacks, which did work in a simpler test script.

PR #369, opened 2024-02-02
Changes for better script online compatibility

estebanfer:
For now it only has a WIP code for the rollback function hook that copies the data from one thread to another, it works but I still have to make it better, add more changes and decide how to name the heap clone functions


src/game_api/state.cpp
```diff
@@ -694,6 +695,76 @@ void init_state_update_hook()
...
+    StateMemory* state_to = reinterpret_cast<StateMemory*>(heap_to + location);
+    heap_clone_event(ON::PRE_CLONE_HEAP, state_from, state_to);
+    // g_heap_clone_trampoline(heap_container_to, heap_container_from, heap_to);
+    // heap_clone_event(ON::POST_CLONE_HEAP, state_from,state_to);
+}
+
+// Original function params: clone_heap(ThreadStorageContainer to, ThreadStorageContainer from)
+// HeapContainer has heap1 and heap2 variables, and some sort of timer, that just increases constantly, I guess to handle the rollback and multi-threaded stuff
+// The rest of what HeapContainer has is unknown for now
+// After writing to a chosen storage from the content of `from->heap1`, sets `to->heap2` to the newly copied thread storage
```
Mr-Auto, line 714: this has to copy more than just state (like prng)
also if you only want callback then figuring out which thread is which from `ThreadStorageContainer` would be enough, then you could just use detour instead of all of this asm

btw. I specially wrote not long ago `patch_and_redirect` function, for exactly this kind of assembly hack, so you don't have to write all the boring jumps there and back and copy game code yourself xd

estebanfer: yeah it also copies more stuff, but honestly I don't know how to handle that well in a callback, maybe make a struct for the whole storage that has prng, state, etc.

The issue with figuring out which thread is, is that I don't really know which thread is going to be the destination thread, at the start of the function you only know that `to->heap2` pointer is gonna get overwritten at the with a pointer to the destination thread.
For getting to which ThreadStorage will be written to, the game code seems to do some lock and unlock stuff, which may affect the other threads, so I decided to not try to figure out to which one will be written to with my own code

I'll look into the `patch_and_redirect` functions

estebanfer: Also, I don't know if I can share some of the ghidra decompiled code to show it

estebanfer:
Should I rename the PRE_CLONE_HEAP event now that there's the other events? (LOAD_STATE, SAVE_STATE)

Dregu:
> Should I rename the PRE_CLONE_HEAP event now that there's the other events? (LOAD_STATE, SAVE_STATE)

Yeah I think the best thing moving forward would be to just call it "state" (so CLONE_STATE or COPY_STATE), grow the existing struct to start at offset 0, renaming it to "State" in the api, absorbing whatever there is before the current StateMemory and possibly after (whatever is part of this heap), making sure the globals always point to the local versions...

estebanfer:
Would `StateMemory` be merged with the current existing `State` struct?

Dregu:
I meant grow the currently existing `StateMemory` struct from the beginning by 0x4a0, and rename it `State`. I know we have a `State` struct already, but parts of it really have nothing to do with the state of anything and is just for initializing the script api. Rest is a random collection of api functions and stupid getter functions to read stuff from just before the `StateMemory`, you know, stuff that would be obsolete when the new `State(Memory)` actually covers that memory.

So

- rename the current `State` to `ScriptAPI` or whatever, keeping the relevant functions, removing the soon obsolete functions
- grow the current `StateMemory` by 0x4a0 and rename it `State`

or

- split the current `State` to `ScriptAPI` and `State`
- put old `StateMemory` inside this new `State`

and

- fill the beginning of new `State` with stuff we know is in it, like the frame count and prng
- organize the functions so new `State` only has functions that deal with that `State`, rest somewhere else

Now I don't exactly know what else is in the beginning of that new `State`, but there's clearly about 0x4a0 bytes of stuff that we should reverse engineer.

Mr-Auto:
> grow the current StateMemory by 0x4a0 and rename it State

> put old StateMemory inside this new State

Those two points are mutually exclusive

There is also a ton of stuff after `StateMemory`, including a lot of the inner structs of `StateMemory`, level gen, liquid physics, but also save data etc.

I don't really see a reason for one giant struct, it will probably just mess with compile time or you will end up fighting the compiler alignments in it.

For the x64dbg plugin i just made one function to get heap address and then you get offset for specific thing via enum

estebanfer:
I was going with the option of growing state, but not I noticed we have some game functions that use StateMemory* with the current size, like `spawn_companion`, `is_inside_active_shop_room`, `update_state`, and some hooks...
So I guess I'm going for the second option by putting StateMemory in State, but would the lua API now use `StateMemory` or `State`?

Dregu:
> I was going with the option of growing state, but not I noticed we have some game functions that use StateMemory* with the current size, like `spawn_companion`, `is_inside_active_shop_room`, `update_state`, and some hooks... So I guess I'm going for the second option by putting StateMemory in State, but would the lua API now use `StateMemory` or `State`?

Well yeah clearly `StateMemory` is its own thing cause it's referenced like that a lot, but also clearly part of a slightly larger thing, so second option makes more sense. We should really figure out what all the stuff in that 0x4a0 even is, but on the lua side things don't really have to change, as we have getters for all the known things in there anyway. Things can be added as properties of `lua["state"]` even if not exactly in `StateMemory` though, if that makes sense.

Even if we end up with just this, it's better than the magic number getters we currently have:

```c++
struct State { // offset 0
  // padding
  PRNG prng;
  uint32_t frame_count;
  // padding
  StateMemory state; // offset 0x4a0
};
```

Still open for suggestions for the naming of those structures, but internally it doesn't really matter that much.

estebanfer:
> Things can be added as properties of lua["state"] even if not exactly in StateMemory though, if that makes sense.

Wouldn't that require doing the same as before, substracting an offset to get the stuff in State?

> Still open for suggestions for the naming of those structures, but internally it doesn't really matter that much.

Having stuff like `state->state` doesn't sound good to me, though I don't have many ideas. Seems that we already have some name for it from before, the `OnHeapPointer` class in `thread_utils.hpp`, but idk why it was named like that

estebanfer:
I made some changes, though I had some doubts, I decided just to do the changes and then get it reviewed

src/game_api/state.hpp
```diff
@@ -316,100 +317,53 @@ struct StateMemory
...
     // Returns the main-thread version of StateMemory*
-    StateMemory* ptr_main() const;
+    static StateMemory* ptr_main();
     // Returns the local-thread version of StateMemory*
-    StateMemory* ptr() const;
-    StateMemory* ptr_local() const;
+    static StateMemory* ptr();
+    static StateMemory* ptr_local();
+    static State* get();
+    static State* get_main();
```
Mr-Auto, line 348: What was wrong with the old way of first getting the state itself?
also i hate this approach of `State::ptr()`, wasn't the whole discussion about the fact that we want `state` to not be just StateMemory but the whole heap? 

now there is this weird dangers casting. Not to mention that all of those functions now should have error check if the returned heap is null (since it can very much happen in the local)

estebanfer: The casting is just the same as before
https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/thread_utils.hpp#L13-L42

estebanfer: should I move this State thing to another PR or something, because the number of changes with this may be too high I guess (?), and also I guess this may need more discussion than the online compatibility stuff I was working on

Dregu: Yeah I guess it turned out bigger and more controversial than I thought, and I wasn't really requesting you to do those changes here and now anyway, this is just where the conversation ended up.

src/game_api/script/lua_backend.cpp
```diff
@@ -36,27 +36,31 @@
...
 LuaBackend::LuaBackend(SoundManager* sound_mgr, LuaConsole* con)
     : lua{get_lua_vm(sound_mgr), sol::create}, vm{acquire_lua_vm(sound_mgr)}, sound_manager{sound_mgr}, console{con}
 {
-    g_state = State::get().ptr_main();
+    g_state = State::get().ptr_local();
+    if (g_state == nullptr)
+    {
+        g_state = State::get().ptr_main();
+    }
+    ScriptState& state = local_state_datas[g_state].state;
```
Mr-Auto, line 51: Isn't backend constructed ASAP in playlunky? then this will always be main state anyway

estebanfer: not sure about when they are constructed, are they when you create a new script in OL?

estebanfer:
I have a local stash of changes to make timers and callbacks started from other callbacks to work online, by moving the callbacks `unordered_map` and other fields from `LuaBackend` to `LocalStateData` and creating custom structs that only get copied when they have changes, so it has many lines of code edited. Should I commit that here or make another PR, to prevent this one from growing too much?

Also, should I make `PRE_COPY_STATE` to be also called from the `savestate` functions? Makes sense to me since people using `PRE_COPY_STATE` probably use it for online mods and wouldn't use the other save callbacks, and also being able to test a part of the rollback without having to open an online game seems be nice

Dregu:
> I have a local stash of changes to make timers and callbacks started from other callbacks to work online, by moving the callbacks `unordered_map` and other fields from `LuaBackend` to `LocalStateData` and creating custom structs that only get copied when they have changes, so it has many lines of code edited. Should I commit that here or make another PR, to prevent this one from growing too much?

Yeah maybe we should do this in smaller parts, but I'll say I have no intention of actually testing any of this online, just to make sure some single player mods still works right.

> Also, should I make `PRE_COPY_STATE` to be also called from the `savestate` functions? Makes sense to me since people using `PRE_COPY_STATE` probably use it for online mods and wouldn't use the other save callbacks, and also being able to test a part of the rollback without having to open an online game seems be nice

Makes sense to me too, even if just for testing.

src/game_api/state.cpp
```diff
@@ -625,6 +626,11 @@ LiquidPhysicsEngine* State::get_correct_liquid_engine(ENT_TYPE liquid_type)
     return nullptr;
 }
 
+size_t State::get_location()
+{
+    return this->location;
+}
```
Mr-Auto, line 632: Simple functions like this should be in header, so that compiler can optimize this call away

Mr-Auto, line 632: Simple functions like this should be in header, so that compiler can optimize this call away

estebanfer:
I removed the PRE_COPY_STATE for now because I don't really know what someone could use it for, so I prefer to leave it out until having something more clear

src/game_api/script/lua_backend.cpp
Mr-Auto: Was thinking more about PL

src/game_api/script/lua_vm.cpp
```diff
@@ -2260,6 +2260,16 @@ end
...
+    lua.script(R"##(
+        function deepcopy_object(obj)
+            if type(obj) ~= 'table' then return obj end
+            local res = {}
+            for k, v in pairs(obj) do res[deepcopy_object(k)] = deepcopy_object(v) end
+            res = setmetatable(res, getmetatable(obj))
+            return res
+        end
+    )##");
+
```
Mr-Auto, line 2272: Are you using this anywhere? or it's just new addition?
does it even work? i know some weird objects like std::vector, or stuff that i did for the global `players` do not list anything with `pairs`

estebanfer: I don't think it would work well with any object other than lua primitives and tables

estebanfer: The part that I don't like is that every time you create a new table, you make new allocations, this means when playing online the object in `state.user_data` and the tables inside it will make many new unnecessary allocations per frame on this function, but I can't think any way of solving this rather than creating a new lua type.
I guess another way would be to not initialize new tables when it already exists, and just copying the primitives in it.
Though, I guess it probably doesn't affect performance much anyway and it's just me wanting to do this very optimized

estebanfer: Oh I didn't read the first part, it is used [here](https://github.com/spelunky-fyi/overlunky/pull/369/files#diff-1162760f250e1dbe80e6b009f3c1201e8b7c46abef51b67df4535f1c169a444fR1891)

estebanfer: I guess I should change [this](https://github.com/spelunky-fyi/overlunky/pull/369/files#diff-d3d0e735e82684d74e7cc32e6a04d7b32540afe19fd45f1c2dd5643752b1fa19R3087) description to say it won't work well with C++ objects then

estebanfer:
On this PR I added `get_location` to `State`, but now there is also `get_state_offset` in `savestate.cpp` which does the same. Which one should be preferred?

src/game_api/script/usertypes/state_lua.cpp
```diff
@@ -414,6 +415,31 @@ void register_usertypes(sol::state& lua)
...
+        auto backend = LuaBackend::get_calling_backend();
+        backend->local_state_datas[&state].user_data = user_data;
+    };
+    auto user_data = sol::property(state_get_user_data, state_set_user_data);
+
+    statememory_type["user_data"] = std::move(user_data);
+    /* StateMemory
+    // user_data
+    // You can store a table (or lua primitive) here and it will store data correctly in online multiplayer, by having a different copy on each state and being copied over when the game does.
+    // Doesn't support recursive tables / cyclic references. Metatables will be transferred by reference instead of being copied
```
Mr-Auto, line 440: > Metatables will be transferred by reference instead of being copied

This is kind of obvious, since you just can't ever do a copy of metatable in lua

Mr-Auto:
> On this PR I added `get_location` to `State`, but now there is also `get_state_offset` in `savestate.cpp` which does the same. Which one should be preferred?

I don't think that matters much. Just pick one

src/game_api/script/usertypes/state_lua.cpp
estebanfer: metatables are just tables, with fields that are accessed by lua when the table is set as a metatable of one, so those are copiable technically, but copying them doesn't make sense for what they are used most, if not all of the time (which is mostly making OOP like methods for tables). Not sure about metatables created by sol2.
https://www.lua.org/pil/13.html

estebanfer:
do you mean to let anyone decide which to use, or to remove one of those?

Mr-Auto:
> do you mean to let anyone decide which to use, or to remove one of those?

Aren't they exactly the same? it doesn't really matter which one we use. So just remove one.
Also this is really not a function that's used often, so it doesn't matter.

Also also, i plan to kill this function anyway, since we can just hardocde the offset 4A0 , it's not a big deal

src/game_api/script/lua_backend.cpp
estebanfer: Yeah but this code is for both PL and OL, also handling what happens in OL makes sense to me

PR #372, opened 2024-02-07
Add SaveState class and events

Dregu:
I guess this copy code works for an arbitrary SaveState class too...

```lua
foo = SaveState:new()
prinspect(foo:get_state().time_level)
set_timeout(function()
  foo:load()
  foo:clear()
  foo = nil
end, 60)
```


src/game_api/savestate.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+struct StateMemory;
+
+class SaveState
+{
+  public:
+    /// Create a new temporary SaveState/clone of the main level state. Unlike save_state slots that are preallocated by the game anyway, these will use 32MiB a pop and aren't freed automatically, so make sure to clear them or reuse the same one to save memory.
```
Mr-Auto, line 8: This sounds kind of ominously, like if you don't clear it, you get memory leak

Dregu: Well I guess the backend could do that for you like for the predefined slots, but we have to be very sure that these are worthless after the level has been unloaded...

Mr-Auto: I was thinking more like, if you just set the variable to nil or overwrite it, it will eventually just clear itself, so the clear is not really useful for people who will make like 2/3 slots, only for stuff like TAS that would make a ton of them

Mr-Auto: Also question: how much did you test the loading save state from previous level? can you just warp to that level and load the saved state or is there something odd that it crashes on?

Dregu: Well the garbage collector does clear them too, it's just not in any hurry to do that automatically...

Dregu: > Also question: how much did you test the loading save state from previous level? can you just warp to that level and load the saved state or is there something odd that it crashes on?

I have tried replaying the same seed and loading an earlier state in an identical level, still crashed. I don't actually understand why or why not that should work, but that's one reason this option is still left there, for someone to find a way to do it.

Dregu: I think I tried both modes, but at least real seeded.

src/game_api/savestate.cpp
```diff
@@ -0,0 +1,121 @@
...
+        return nullptr;
+    return reinterpret_cast<StateMemory*>(addr + 0x4a0);
+}
+
+void SaveState::load()
+{
+    if (!addr)
+        return;
+    size_t to = (size_t)(State::get().ptr_main()) - 0x4a0;
+    auto state = reinterpret_cast<StateMemory*>(addr + 0x4a0);
```
Mr-Auto, line 96: We actually have a pattern to get this offset `0x4a0`, but if you want to keep it as raw value, can you make it into a global constexpr?
it is also in the State::get() as default (if the pattern would fail for some reason)

Dregu: The best would probably be if someone made a struct of the whole heap thingy with prng, state etc in it so we wouldn't need this.

PR #378, opened 2024-03-19
Skip script chance logic providers if their spawn chance is zero

estebanfer:
HDMod was having big issues with level loading times, turns out it was due to script procedural chances still being run when they were set to zero, with this change, average level loading changes got ~5x times better on it. It should be around the same on most small scripts.

Also, should I make the same change to community chances? I guess it would make sense

### DIFF
--- a/src/game_api/level_api.cpp
+++ b/src/game_api/level_api.cpp
@@ -1279,9 +1279,10 @@ bool handle_chance(SpawnInfo* spawn_info)
 
     uint8_t layer = 0;
     auto* layer_ptr = State::get().layer(layer);
+    LevelGenSystem* level_gen = State::get().ptr()->level_gen;
     for (const CommunityChance& community_chance : g_community_chances)
     {
-        if (community_chance.test_func(community_chance, spawn_info->x, spawn_info->y, layer_ptr))
+        if (level_gen->get_procedural_spawn_chance(community_chance.chance_id) != 0 && community_chance.test_func(community_chance, spawn_info->x, spawn_info->y, layer_ptr))
         {
             if (g_test_chance(&level_gen_data, community_chance.chance_id))
             {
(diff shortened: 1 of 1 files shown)


estebanfer:
Any opinions or updates on if this is gonna get merged?
Should I not use a different vector to store the enabled chances? Maybe just checking the chance in `handle_chance` and not running the `provider.is_valid` if the chance is zero would be performant enough?

Dregu:
> Any opinions or updates on if this is gonna get merged?

Oh right, in my headcanon I had twisted your earlier statement about community chances as something you still want to add to this, and maybe we should if that's also a big performance issue.

> Should I not use a different vector to store the enabled chances? Maybe just checking the chance in `handle_chance` and not running the `provider.is_valid` if the chance is zero would be performant enough?

Well if this was a big performance issue, whatever sounds like the most performant choice, it's not that much memory wasted. I'm not very familiar with this part of the api and can't really think what kind of a hit it would have compared to this.

estebanfer:
I haven't tested how the community chances affect the performance, probably less than script chances since it's with compiled code. But still, running checks to verify a spot is valid for spawning a thing when it's definitely not going to spawn anywhere, seems like a waste...

estebanfer:
Also, a very weird case where this optimization could cause a bug, is if someone decides to change a chance from 0 to higher, inside an `is_valid` or `do_spawn` function by using the `PostRoomGenerationContext:new()` to access the chance. Though seems very specific and you could already expect not everything to work by using `PostRoomGenerationContext` outside of the callback made for it.
To me this maybe bug seems seems justifiable due to my statement right above, idk what you think.

Mr-Auto:
> Also, a very weird case where this optimization could cause a bug, is if someone decides to change a chance from 0 to higher, inside an `is_valid` or `do_spawn` function by using the `PostRoomGenerationContext:new()` to access the chance. Though seems very specific and you could already expect not everything to work by using `PostRoomGenerationContext` outside of the callback made for it. To me this maybe bug seems seems justifiable due to my statement right above, idk what you think.

there is `sol::no_constructor` if you don't want people to be able to construct object

estebanfer:
there are some valid use cases of constructing it tho, like `cells_interlinked.lua` has to use it since it stops the `POST_ROOM_GENERATION` callback from even happening

Dregu:
Yeah I specifically added the constructor for it since we put too many good functions behind it, but it's still not a very convenient way to use those functions outside the callback. There were a lot less callbacks when it was made.

estebanfer:
I tried not using a vector to store enabled chances, and just check if the chance was not zero before the valid checks... Turns out it was actually more performant than the one using the vector, so I deleted the other commit and made a new one, this should be the final one
Also, adding the optimization to community chances made a change of time spent on procedurals spawn, from previously ~0.01s to ~0.0015s (tested on dwellings), I guess more than x5 faster, but very small change

PR #379, opened 2024-03-28
StateMemory and State structs changes

SereneRuby12:
Some StateMemory and State changes, discussion started in https://github.com/spelunky-fyi/overlunky/pull/369#issuecomment-1985325199

### DIFF
--- a/src/game_api/mod_api.cpp
+++ b/src/game_api/mod_api.cpp
@@ -1,0 +1,455 @@
+#include <detours.h> // for DetourAttach, DetourTransactionBegin
+
+#include "bucket.hpp"         // for Bucket
+#include "entities_chars.hpp" // for Player
+#include "game_manager.hpp"
+#include "game_patches.hpp"
+#include "items.hpp"
+#include "memory.hpp"
+#include "mod_api.hpp"
+#include "movable.hpp"
+#include "movable_behavior.hpp"
+#include "script/events.hpp"                     // for pre_entity_instagib
+#include "script/lua_vm.hpp"                     // for get_lua_vm
+#include "script/usertypes/theme_vtable_lua.hpp" // for NThemeVTables
+#include "search.hpp"
+#include "spawn_api.hpp"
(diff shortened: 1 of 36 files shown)


src/game_api/mod_api.cpp
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+void hook_godmode_functions()
+{
+    static bool functions_hooked = false;
+    if (!functions_hooked)
+    {
+        auto memory = Memory::get();
+        auto addr_damage = memory.at_exe(get_virtual_function_address(VTABLE_OFFSET::CHAR_ANA_SPELUNKY, 48));
```
Mr-Auto, line 125: I know you didn't write this code, but wondering why this is done via Detour probably better to do it via `hook_vtable` ?

src/game_api/rpc.cpp
```diff
@@ -260,7 +258,7 @@ std::vector<Player*> get_players(StateMemory* state)
 {
```
Mr-Auto, line 258: This function should probably be just member of `StateMemory`

src/game_api/rpc.cpp
```diff
@@ -448,20 +444,20 @@ void flip_entity(uint32_t uid)
 
 void set_camera_position(float cx, float cy)
 {
-    auto& state = State::get();
-    state.set_camera_position(cx, cy);
+    StateMemory* state = State::ptr();
+    state->set_camera_position(cx, cy);
 }
 
 void warp(uint8_t world, uint8_t level, uint8_t theme)
```
Mr-Auto, line 451: All of those simple functions that just get StateMemory and return some value from it or based on it, should probably be just member functions of the StateMemory. Then in `lua_vm` just make lambda.

Mr-Auto:
Would you mind if i try and make my own version of this PR ?

estebanfer:
Maybe it would be even better if you're the one that makes this PR tbh

PR #382, opened 2024-05-06
Online scripts compatibility stuff pt. 2

estebanfer:
Rollback with LuaBackend's timeouts, callbacks and script state, perhaps I should try using the `custom_allocator`, for the first try I decided to just copy stuff manually when the game does too, and seems to work fine, and I'm not sure which one would be better.
This is the continuation of #369


Mr-Auto:
So this is all the changes in previous PR + new stuff? 
that will be hard to look thru what's new about it

estebanfer:
The idea is that the previos PR gets merged first, then the changes of this one will be more clear

src/game_api/online_util.hpp
```diff
@@ -0,0 +1,116 @@
+// Classes with only the functions that were currently used, more can be added if needed
```
Mr-Auto, line 1: `#pragma once`

Mr-Auto (approved):

src/game_api/online_util.hpp
estebanfer: done

SereneRuby12:
The issue was that there may be desyncs due to callback execution order if a player does something that starts more callbacks before joining multiplayer.
E.g: One player plays a run, which starts many timeouts through it. Then when they join in a multiplayer game, the timeout id would be different to other players timeout ids, so they could end up executing in different order causing random unexpected desyncs. I haven't actually tested if this happens or not, but I would say it's likely that this could happen.
It applies to any other callback since they're unordered.

I couldn't take a decision on how to solve this, idk if using an ordered map could cause performance issues on callback heavy mods.

Mr-Auto:
> idk if using an ordered map could cause performance issues on big callback heavy mods.

Pretty sure even if you use vector it would be fine.
My call would be, do it however you thing is fine, if it brings problems we can figure out better storage solution (like the ordered hash map i suggested)

src/game_api/online_util.hpp
```diff
@@ -0,0 +1,117 @@
...
+        {
+            unordered_map::operator=(other);
+            edit_num = other.edit_num;
+        }
+        return *this;
+    }
+    inline void increase_edit_num()
+    {
+        edit_num++;
+    }
```
Mr-Auto: Wait, why are those method public? I don't think they are even used?

PR #383, opened 2024-05-17
Fix damage virtual in the api doc

Mr-Auto:


estebanfer:
Does this break compatibility with the previos `movable:damage()` function when called from lua?

Mr-Auto:
> Does this break compatibility with the previos `movable:damage()` function when called from lua?

No, you can still call it like:
```lua
entity:damage(damage_dealer_uid, damage_amount, stun_time, velocity_x, velocity_y)
entity:damage(damage_dealer_uid, damage_amount, stun_time, velocity_x, velocity_y, iframes)
entity:damage(damage_dealer, damage_amount, damage_flags, velocity, unknown_damage_phase, stun_amount, iframes, unknown_is_final)
```
unless i missed something? can't test the current overlunky version right now

estebanfer:
Ah I see, yeah it should work I think

src/game_api/movable.hpp
```diff
@@ -92,11 +92,16 @@ class Movable : public Entity
...
 
-    /// Damage the movable by the specified amount, stuns and gives it invincibility for the specified amount of frames and applies the velocities
-    /// Returns: true if entity was affected, damage_dealer should break etc. false if the event should be ignored by damage_dealer?
-    bool damage(uint32_t damage_dealer_uid, int8_t damage_amount, uint16_t stun_time, float velocity_x, float velocity_y, uint8_t iframes);
-    // the original damage function was added to the API without the iframes param, but for backwards compatibility we preserve the broken one
-    bool broken_damage(uint32_t damage_dealer_uid, int8_t damage_amount, uint16_t stun_time, float velocity_x, float velocity_y);
+    /// NoDoc
+    bool broken_damage(uint32_t damage_dealer_uid, int8_t damage_amount, uint16_t stun_time, float velocity_x, float velocity_y, std::optional<uint8_t> iframes)
+    {
+        auto dealer = get_entity_ptr(damage_dealer_uid);
```
Mr-Auto, line 98: One think i worry about, this should check if the uid is `-1` vs `get_entity_ptr` returning `nullptr`
As the first one is probably intended, when the second might be not

PR #387, opened 2024-06-26
Use Vec2 more, some small issues fix, more unknown stuff about camera and Hud

Mr-Auto:


src/game_api/containers/custom_allocator.cpp
```diff
@@ -12,12 +12,12 @@ using CustomFreeFun = void*(void*, void*);
 void* custom_malloc(std::size_t size)
 {
     static CustomMallocFun* _malloc = (CustomMallocFun*)get_address("custom_malloc"sv);
-    static void* _alloc_base = OnHeapPointer<void>(*(size_t*)get_address("malloc_base"sv)).decode();
+    static void* _alloc_base = OnHeapPointer<void>(*(size_t*)get_address("malloc_base"sv)).decode(); // probably should be decode_local
```
Mr-Auto, line 15: If i understand all about the custom allocator, it allocates the stuff in memory that is copied over when doing save state (?)

So then this definitely should be local, but maybe I'm wrong?

src/game_api/state.cpp
```diff
@@ -539,7 +539,7 @@ void State::set_seed(uint32_t seed)
 SaveData* State::savedata()
 {
     auto gm = get_game_manager();
-    return gm->save_related->savedata.decode();
+    return gm->save_related->savedata.decode(); // wondering if it matters if it's local or not?
```
Mr-Auto, line 542: It probably kind of is? if someone unlocks something in online and then immediately gets re-sync to earlier moment

src/game_api/state_structs.hpp
```diff
@@ -137,17 +138,30 @@ struct Camera
...
-    int32_t focused_entity_uid; // if set to -1, you have free control over camera focus through focus_x, focus_y
-    uint32_t freeze_timer;      // if > 0, disables camera movement for this amount of frames
-    uint32_t unknown4;
+    uint8_t padding1[3];
+    /// if set to -1, you have free control over camera focus through focus_x, focus_y
+    int32_t focused_entity_uid;
+    /// amount of frames to freeze camera in place and move to the peek_layer
+    /// during the peek you can freely set camera position no matter if focused_entity_uid is set to -1 or not
+    uint32_t peek_timer;
+    uint8_t peek_layer;
```
Mr-Auto, line 147: Better name suggestions?

src/game_api/containers/custom_allocator.cpp
estebanfer: Yes it should be local
https://github.com/spelunky-fyi/overlunky/pull/369/commits/f93ee54120bc70b1108a81a80c1c03b79c101724

src/game_api/state.cpp
estebanfer: I guess using the local state is always a good thing to do

Mr-Auto: I think i won't change this thou, it's not super important. Maybe you can add that to your online PR stuff

Mr-Auto:
I guess i mark this as ready as well, if there are no objections to that variable name change it can be merged

Dregu (changes requested):

Dregu (approved):

PR #389, opened 2024-06-30
Allow creating color from html color names

Mr-Auto:


Mr-Auto:
Again tried to implement HSV/HSL conversions with no luck

PR #390, opened 2024-07-01
fix issue with returning self reference not counting as lua owned object

Mr-Auto:
Draft since i did some template hell and maybe someone can suggest some improvements or better idea completely


Mr-Auto:
Will look thru more types like Color... etc. later

docs/game_data/spel2.lua
```diff
@@ -2446,11 +2445,9 @@ function PRNG:random(min, max) end
     ---@field texture_num integer
     ---@field get_entity fun(self): Entity
     ---@field set_normal_map_texture fun(self, texture_id: TEXTURE): boolean @Sets second_texture to the texture specified, then sets third_texture to SHINE_0 and texture_num to 3. You still have to change shader to 30 to render with normal map (same as COG normal maps)
-    ---@field get_second_texture TEXTURE?
-    ---@field get_third_texture TEXTURE?
```
Mr-Auto, line 2450: I have no idea why is this happening, it should now be recognized as a function, but for some reason it does not print it

PR #392, opened 2024-08-03
Add get_liquids_at function and small stuff

estebanfer:
I added the `ENTITY_MASK` enum, I didn't if name it `MASK` because sounds too much of a generic name, but if someone thinks it's fine then I'll change it

### DIFF
--- a/examples/liquids.lua
+++ b/examples/liquids.lua
@@ -1,3 +1,7 @@
+meta.name = "Liquids by grid"
+meta.version = "1.0"
+meta.description = "Shows the amounts of liquids by grid, using the `get_liquids_at` function"
+
 register_option_bool("only_draw_at_mouse", "Testing: Draw only liquid num at mouse position", "", false)
 register_option_bool("draw_grid_lines", "Draw grid lines", "", true)
 register_option_bool("draw_text_shadow", "Draw text shadow", "", true)
@@ -39,8 +43,7 @@ end
 ---@param x number
 ---@param y number
 local function draw_liquids_at(ctx, x, y)
-	local water = get_liquids_at(MASK.WATER, x, y, LAYER.PLAYER)
-	local lava = get_liquids_at(MASK.LAVA, x, y, LAYER.PLAYER)
+	local water, lava = get_liquids_at(x, y, LAYER.PLAYER)
 	if water == 0 and lava == 0 then return end
 
 	local spacing <const> = 0.0135 / last_zoom
--- a/examples/liquids.lua
+++ b/examples/liquids.lua
@@ -1,0 +1,120 @@
+register_option_bool("only_draw_at_mouse", "Testing: Draw only liquid num at mouse position", "", false)
+register_option_bool("draw_grid_lines", "Draw grid lines", "", true)
+register_option_bool("draw_text_shadow", "Draw text shadow", "", true)
+
+local function get_camera_bounds_grid()
+	local left, top = game_position(-1, 1)
+	local right, bottom = game_position(1, -1)
+	left, top, right, bottom = math.floor(left - 0.5), math.ceil(top + 0.5), math.ceil(right + 0.5), math.floor(bottom - 0.5)
+	return left, top, right, bottom
+end
+
+local water_color = Color:new(0, 0.1, 1.0, 1.0)
+local lava_color = Color:new(1.0, 0.1, 0.0, 1.0)
+local black = Color:black()
+
+---@type table<integer, TextRenderingInfo>
+local num_text_renders = {}
+local last_zoom = get_zoom_level()
+
+-- Generate a text for every possible number, to optimize when there's many numbers to render
+local function reload_num_text_renders(zoom)
+	for i = 0, 12 do
+		num_text_renders[i] = TextRenderingInfo:new(tostring(i), 0.0081 / zoom, 0.0081 / zoom, VANILLA_TEXT_ALIGNMENT.CENTER, VANILLA_FONT_STYLE.NORMAL)
+	end
+end
+reload_num_text_renders(last_zoom)
+
+---@param ctx VanillaRenderContext
+---@param text_render TextRenderingInfo
+local function draw_text_shadow(ctx, text_render)
+	local offset <const> = 0.027 / last_zoom
+	local saved_x, saved_y = text_render.x, text_render.y
+	text_render.x, text_render.y = text_render.x + offset, text_render.y - offset
+	ctx:draw_text(text_render, black)
+	text_render.x, text_render.y = saved_x, saved_y
+end
+
+---@param ctx VanillaRenderContext
+---@param x number
+---@param y number
+local function draw_liquids_at(ctx, x, y)
+	local water = get_liquids_at(MASK.WATER, x, y, LAYER.PLAYER)
+	local lava = get_liquids_at(MASK.LAVA, x, y, LAYER.PLAYER)
+	if water == 0 and lava == 0 then return end
+
+	local spacing <const> = 0.0135 / last_zoom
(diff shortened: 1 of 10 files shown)


src/game_api/aliases.hpp
```diff
@@ -190,3 +190,25 @@ enum class PAUSE_SCREEN : int64_t
...
+    ACTIVEFLOOR = 0x80,
+    FLOOR = 0x100,
+    DECORATION = 0x200,
+    BG = 0x400,
+    SHADOW = 0x800,
+    LOGICAL = 0x1000,
+    WATER = 0x2000,
+    LAVA = 0x4000,
+    LIQUID = 0x6000,
+    NONE = 0x0,
```
Mr-Auto, line 212: We use 0 as ANY, not NONE. This can be confusing, maybe just add both?

estebanfer: Perhaps both, it was weird to do something like `(mask & ENTITY_MASK::WATER) == ENTITY_MASK::ANY`

Mr-Auto: I would do `== 0`  in that case. Or the inverse `!= ENTITY_MASK::WATER`
Using mask to test for another mask feels weird
(and yes, i know it it's not allowed by default with enum class in C++)

estebanfer: it's kinda annoying that it's not possible to just do `if (mask & ENTITY_MASK::WATER)` and you have to do either `(mask & ENTITY_MASK::WATER) == ENTITY_MASK::WATER` or `static_cast<uint32_t>(mask & ENTITY_MASK::WATER)`

estebanfer: Would this be fine in `aliases.hpp`?
```cpp
template <class T>
bool test_flags(T flags, T test_flags)
{
    return (flags & test_flags) == test_flags;
}
```

Mr-Auto:
Btw. I'm still trying to figure out why would anyone want to know number of liquid entity/particles in a tile?
what's the use?

estebanfer:
The array is how the game mostly checks for liquid collisions (eg. `Entity:is_in_liquid()` uses it), it makes sense to also use it for mods, and not be checking for if every blob of liquid is colliding a hitbox

src/game_api/aliases.hpp
Dregu: Whatever you think makes this enum easier to use and compare is fine.

Mr-Auto: I also found another solution:
```cpp
namespace ENTITY_MASK
{
enum
{
    WATER = 0x2000,
    LAVA = 0x4000,
// etc.
}
}
```

estebanfer: hmm yeah it works, and since what was being used before were raw integers, its still an improvement

estebanfer: oh, it doesn't allow to use ENTITY_MASK as a type then

estebanfer: I'm just gonna leave it like that, at least for now, as the rest of flags enums are

Mr-Auto: > oh, it doesn't allow to use ENTITY_MASK as a type then

The `namespace` solution? it does, just have to give the enum itself name as well, it will look silly thou, with the type being `ENTITY_MASK::ENUM_NAME` and the enums only requiring the namespace.

Mr-Auto: Also: I'm wondering if there is some compiler flag that would force the use of `ENUM_NAME::VAR` for the C style enums 🤔

estebanfer: like this?
```cpp
namespace ENTITY_MASK
{
enum TYPE : uint32_t
{ ... }}
```
```cpp
uint8_t get_liquids_at(ENTITY_MASK::TYPE liquid_mask, float x, float y, LAYER layer)
```

Mr-Auto: yeah, thou I'm not suggesting that this is good solution, just one that exist

estebanfer: I think what I did now works well enough

estebanfer:
This is be ready for review/merge now

src/game_api/rpc.cpp
```diff
@@ -1354,6 +1355,33 @@ void add_entity_to_liquid_collision(uint32_t uid, bool add)
...
+{
+    uint8_t actual_layer = enum_to_layer(layer);
+    LiquidPhysics* liquid_physics = State::get().ptr()->liquid_physics;
+    if (actual_layer != get_liquid_layer())
+        return 0;
+    if (y >= 125.5 || y < 0 || x >= 85.5 || x < 0) // Game uses `>` instead of `>=`, resulting in it being able to read out of bounds of the array, but it never happens without OL because of the border blocks
+        return 0;
+
+    uint32_t ix = static_cast<int>((x + 0.5) / 0.3333333);
+    uint32_t iy = static_cast<int>((y + 0.5) / 0.3333333);
```
Mr-Auto, line 1368: I don't think it changes much with adding the `=` mark, since it's hard for floats to be exact equal, should probably first convert to int, maybe also then you could use the `g_level_max_x` and `g_level_max_y` values (from layer.hpp)

Also feel like on every case we convert float to int slightly different

estebanfer: oh, like checking array bounds after converting to int? makes sense

estebanfer: should be solved, idk if I do mark this as solved or the reviewer is the one that should do it

Mr-Auto: Don't think it matters. As PR creator i think you always have the ability to "resolve", so it would kind of suggest that the creator should do it?. 

Anyway, I think the rule should be: either reviewer merges the PR or allows for it to be merged. And in both cases it's up to the reviewer to make sure all the issues have been resolved

Mr-Auto: Don't think it matters. As PR creator i think you always have the ability to "resolve", so it would kind of suggest that the creator should do it?. 

Anyway, I think the rule should be: either reviewer merges the PR or allows for it to be merged. And in both cases it's up to the reviewer to make sure all the issues have been resolved

Mr-Auto:
Hmmm now that i think about this, instead of using `ENTITY_MASK` for this, which contains much more then just the two liquids, you could just return tuple with both, water and lava. And whoever uses this can decide to use one or both in whatever way they want to

examples/liquids.lua
```diff
@@ -0,0 +1,120 @@
+register_option_bool("only_draw_at_mouse", "Testing: Draw only liquid num at mouse position", "", false)
```
Mr-Auto, line 1: We usually add the `meta` information for the examples, since they are loaded into the overlunky automatically and shown in the scripts tab with the name and have small description on what they are

estebanfer:
> Hmmm now that i think about this, instead of using `ENTITY_MASK` for this, which contains much more then just the two liquids, you could just return tuple with both, water and lava. And whoever uses this can decide to use one or both in whatever way they want to

It would have been better months ago when I asked about it... but well that makes sense

examples/liquids.lua
estebanfer: oopsie I forgot

Mr-Auto (approved):
All looks good
last note: in `get_liquids_at` you didn't specify the literal values as floats, so the math is done with doubles

Not a big deal, probably done many times all over the API anyway

estebanfer:
> last note: in get_liquids_at you didn't specify the literal values as floats, so the math is done with doubles

oh, you're right... I'm pretty perfectionist so I'll fix that anyway, it's just adding an `f` at the end of the literals :cat2:

PR #396, opened 2024-09-11
Fixes to require and loadlib

Dregu:
It was recently brought back to my attention that loading DLLs with `require` simply didn't work at all, and `loadlib` could only load them from the game root.

- Fixed loading DLLs with `require` from the mod folder
- Fixed loading DLLs with `package.loadlib` from the mod folder
- Fixed silly 3rd party scripts using `require` on built-in libraries just setting them to nil

This should allow things like the [archipelago mod](https://spelunky.fyi/mods/m/spelunky-2-archipelago/) to simply `require "apclientpp"` when `apclientpp.dll` is vendored with the mod, instead of all the extra steps when installing it.

The following also seem to work for example, when a complete luasocket build is provided to the mod dir:

```lua
-- Loads socket.lua, which later requires the actual core.dll with require("socket.core"), which all seems to work out
local socket = require("socket")
local mime = package.loadlib("mime/core.dll", "luaopen_mime_core")()
```

### DIFF
--- a/src/game_api/script/lua_require.cpp
+++ b/src/game_api/script/lua_require.cpp
@@ -25,23 +25,29 @@ void register_custom_require(sol::state& lua)
     lua.add_package_loader(custom_loader);
 
     lua["__require"] = lua["require"];
+    lua["__loadlib"] = lua["package"]["loadlib"];
 
     /// Custom implementation to trick Lua into allowing to `require 'lib.module'` more than once given it was called from a different source
     lua["require"] = custom_require;
+
+    lua["package"]["loadlib"] = custom_loadlib;
 }
 sol::object custom_require(std::string path)
 {
+    static sol::state& lua = get_lua_vm();
+
+    if (path == "io" || path == "os" || path == "math" || path == "string" || path == "table" || path == "coroutine" || path == "package")
+        return lua[path];
+
     // Turn module into a real path
     {
-        if (path.ends_with(".lua"))
+        if (path.ends_with(".lua") || path.ends_with(".dll"))
         {
             path = path.substr(0, path.size() - 4);
         }
         std::replace(path.begin(), path.end(), '.', '/');
         std::replace(path.begin(), path.end(), ':', '.');
     }
-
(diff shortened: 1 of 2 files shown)


examples/udp.lua
```diff
@@ -0,0 +1,82 @@
...
+    host = "127.0.0.1",
+    port = 0
+}
+
+function init()
+    deinit()
+
+    server = {
+        -- Save messages to queue instead of printing directly, cause that doesn't work from the server thread
+        queue = {},
```
Mr-Auto, line 18: Have you though about making safe version of `udp_listen`? so people don't have to worry about it being on different thread

Mr-Auto: Maybe instead of callback, having a function that you can use to request the current buffer of messages

Dregu: I have actually made exactly that locally already, but haven't pushed it cause there's also a bunch of tcp stuff that doesn't work at all.

Dregu:
Yeah sorry, I nuked my whole windows installation in frustration in january and still haven't thought about getting it back in some vm form, literally can't do anything about my windows projects 😅. Maybe in the winter...

I'm not even sure if I still have the code I talked about, but anyway you are allowed to fix this and merge it. My memory is hazy, but

>     1. That `static` buffer

Silly me.

>     2. Not sure if the `.dll` loading is unsafe only

Pretty sure all of loadlib and require(.dll) are behind unsafe checks in our custom versions of the functions.

Mr-Auto:
ok, looked thru this code again. I don't really like the `UdpServer` part, creating it uses `std::move` of the socket, which should consider the socket no longer valid, yet it is still accessed. 
And it feels like there could be situation where the `UdpServer` is destroyed while the other thread still does stuff, potentially still doing stuff with the pointer to the now invalid `UdpServer`. 

Or when calling `close` while the other thread tires to do `recv_from` or `send_to`, (this probably will never crash due to how simple operations those are and checks that will kind of be thread safe plus closing socket taking some time etc.)

I have some ideas on how to make this better, but feel like i should make separate PR for this

PR #397, opened 2024-09-11
Entity vtable fixes and new stuff

Mr-Auto:


src/game_api/sound_manager.hpp
```diff
@@ -260,6 +260,9 @@ struct SoundInfo
     std::string sound_name; // not 100% sure if it's standard
 };
 
+// there is actually base class that consists of up to the left/right channel, used for the sfx
+// then normal sounds that you can find in entities, screens itp. have up to the paddings
+// finally music in GameManages has the extra bool in BackgroundSound
```
Mr-Auto, line 265: Should i split this? (if yes, would love some name recommendations)

src/game_api/vtable_hook.hpp
```diff
@@ -55,7 +55,7 @@ struct VTableDetour<RetT(ClassT*, ArgsT...), Index>
                 s_Functions[self](self, args..., s_Originals[vtable]);
                 return;
             }
-            s_Originals[vtable](self, std::move(args)...);
+            s_Originals[vtable](self, args...);
```
Mr-Auto, line 58: This is kind of bad, but also all the parameters in virtuals are just simple values and pointers, so there shouldn't be any big optimization issues.
Had to do this for reference parameters that won't work with std::move

src/game_api/rpc.cpp
```diff
@@ -577,14 +573,25 @@ void pick_up(uint32_t who_uid, uint32_t what_uid)
...
+    if (what_uid.has_value()) // should we handle what_uid = -1 the same way?
     {
-        ent->drop(item);
+        auto item = get_entity_ptr(what_uid.value());
+        if (item == nullptr)
+            return;
+        if (item->overlay != mov && mov->holding_uid == what_uid)
+            return;
     }
+    mov->drop();
```
Mr-Auto, line 594: Asked this on discord but didn't get replay
this changes the behavior slightly, but makes the most sense with the parameters given

src/game_api/entity.cpp
```diff
@@ -33,9 +33,9 @@ using namespace std::chrono_literals;
 void Entity::teleport(float dx, float dy, bool s, float vx, float vy, bool snap)
 {
     if (overlay)
-        overlay->remove_item(uid);
-    overlay = NULL;
-    auto topmost = topmost_mount();
+        overlay->remove_item(this, false);
+
+    auto topmost = topmost_mount(); // we just detached from overlay, why do we check this?
```
Mr-Auto, line 38: wondering if this is just leftover from something else?

Dregu: Maybe some old crap the UI used to need, maybe the remove_item wasn't there or maybe we wanted to really make sure we're definitely moving the topmost thing even though it should be detached already 🤷🏻‍♂️. These are like week 1 OL functions.

src/game_api/rpc.cpp
Dregu: So, did we think this virtual could be used to remove any entity from the items, but it actually just drops whatever is held when the uid is valid? Or did it have a different implementation that actually worked like it says, and was then simplified to a more broken one? Seems like an ok compromise anyway.

src/game_api/entities_items.hpp
```diff
@@ -739,7 +756,8 @@ class RollingItem : public Purchasable
   public:
     float roll_speed; // only positive numbers
 
-    virtual void on_purchase(Entity* who, int32_t unknown) = 0; // give you the powerup if you buy it
+    /// Skip this function for item to be unpickable
+    virtual void pickup(Entity* who, bool) = 0;
```
Mr-Auto, line 760: Better name suggestions? (since we already have `pick_up` function for movable

src/game_api/entity.cpp
Mr-Auto: I will leave it as is but move it to `ui_util` since it's only used for UI

estebanfer (commented):
Most of it is great! Just some small things to fix

src/game_api/entities_items.hpp
Mr-Auto: no, this is only for all the `ENT_TYPE_ITEM_PICKUP_*`, it applies the effect of giving you ropes, power up etc.
I mostly want it as hookable function, for stuff like custom pickups

src/game_api/script/usertypes/vtables_lua.cpp
```diff
@@ -18,58 +26,109 @@ void register_usertypes(sol::state& lua)
...
+        // v73
+        VTableEntry<"is_powerup_capable", 74, MemFun<&Movable::is_powerup_capable>>, // dunno if called by the game, might be inlined?
+        VTableEntry<"initialize", 75, MemFun<&Movable::initialize>>,
+        // check_is_falling
+        // v77
+        VTableEntry<"process_input", 78, MemFun<&Movable::process_input>>,
+        // post_collision_damage_related?
+        VTableEntry<"picked_up", 80, MemFun<&Movable::on_picked_up>>,
+        VTableEntry<"release", 81, MemFun<&Movable::on_release>>,
+        VTableEntry<"generate_fall_poof_particles", 82, MemFun<&Movable::generate_fall_poof_particles>>,
```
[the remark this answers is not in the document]
Mr-Auto: so maybe get rid of the "poof" part? or you have better name suggestion?

src/game_api/entities_monsters.hpp
```diff
@@ -384,7 +378,7 @@ class Lavamander : public Monster
     uint16_t jump_pause_timer; // jump pause when cool; runs continuously when hot
     uint8_t lava_detection_timer;
     bool is_hot;
-    /// 0 - didnt_saw_player, 1 - saw_player, 2 - spited_lava; probably used so he won't spit imminently after seeing the player
+    /// 0 - didn't saw player, 1 - saw player, 2 - spited lava; probably used so he won't spit imminently after seeing the player
```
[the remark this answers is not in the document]
Mr-Auto: English is so stupid sometimes

src/game_api/movable.hpp
```diff
@@ -143,88 +138,80 @@ class Movable : public Entity
...
-    virtual void pick_up(Entity* entity_to_pick_up) = 0;         // 67
-    virtual void on_picked_up_by(Entity* entity_picking_up) = 0; // 68
+    virtual void check_out_of_bounds() = 0;               // 58, kills with the 'still falling' death cause, is called for any item/fx/mount/monster/player
+    virtual void set_standing_on(int32_t entity_uid) = 0; // 59
+    virtual Entity* standing_on() = 0;                    // 60
+    virtual bool on_stomped_on_by(Entity* stomper) = 0;   // 61
+    virtual void on_thrown_by(Entity* thrower) = 0;       // 62, implemented for special cases like hired hand (player with ai_func), horned lizard...
+    /// Entities must be of the same type!
+    virtual void copy_extra_info(Entity* clone, int32_t some_entity_uid) = 0; // 63, some_entity_uid - only used for CHAR_ entities, related to hired hand chain
+    virtual uint32_t get_type_id() = 0;                                       // 64, dunno what for, implemented solely that ITEM_EXCALIBUR can return ITEM_BROKENEXCALIBUR instead
```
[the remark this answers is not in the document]
Mr-Auto: Maybe, if anyone wants to test it (if it's used for anything else) and expose in future, go ahead

src/game_api/script/usertypes/vtables_lua.cpp
```diff
@@ -18,58 +26,109 @@ void register_usertypes(sol::state& lua)
...
+        // damage_entity
+        // v73
+        VTableEntry<"is_powerup_capable", 74, MemFun<&Movable::is_powerup_capable>>, // dunno if called by the game, might be inlined?
+        VTableEntry<"initialize", 75, MemFun<&Movable::initialize>>,
+        // check_is_falling
+        // v77
+        VTableEntry<"process_input", 78, MemFun<&Movable::process_input>>,
+        // post_collision_damage_related?
+        VTableEntry<"picked_up", 80, MemFun<&Movable::on_picked_up>>,
+        VTableEntry<"release", 81, MemFun<&Movable::on_release>>,
```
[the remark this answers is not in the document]
Mr-Auto: The comment is meaningless, it only means those entities have this virtual function implemented, but this function works for all entities, it is called on throw or drop etc.

src/game_api/entities_items.hpp
estebanfer: what about uh, `get_powerup`, `give_powerup`, `grant_powerup`, `pick_powerup`, `consume_powerup`, `apply_powerup`?

src/game_api/movable.hpp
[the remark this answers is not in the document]
estebanfer: So I guess nothing for this PR

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
estebanfer: `generate_fall_fx` or `generate_fall_effects`? Maybe add `floor_` before `fall`

src/game_api/entities_items.hpp
Mr-Auto: I would not consider bomb bag a powerup, or like, not in current naming scheme that we or the game established
all of those suggestions sounds more like stand-alone functions rather then method on an object

Mr-Auto: thou, if i had to i would probably choose `pick_powerup`

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
estebanfer: Oh so that comment is wrong, it says only for hired hands and lava pots

src/game_api/movable.hpp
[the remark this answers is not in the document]
Mr-Auto: I just want to be done with this one, there is still work to do on exposing more, like the mount sounds that can be useful for making custom ones etc.

src/game_api/sound_manager.hpp
estebanfer: I guess it would be nice if it means more of those sound types exposed. But I don't really know about name, since now I guess we only have to use a name without altering the SoundMeta one...

Mr-Auto: You can edit the name, as long as there is no `new` constructor exposed to lua, no one should really care about the metatable name

src/game_api/entities_items.hpp
estebanfer: > I would not consider bomb bag a powerup, or like, not in current naming scheme that we or the game established

I don't really see it like that from what I've seen, there's no POWERUP_BOMB or POWERUP_ROPE because there is just no need of such entities, and not because those aren't powerups (also I read in wikipedia, powerups like 1up or ammo are actually named powerup in there)

> all of those suggestions sounds more like stand-alone functions rather then method on an object

Fair, I would be fine with any other name as long as it's distinguishable enough from the `pick_up` functions that's just grabbing something your hands. Though for me it seems fine to me to see some code like `pickup:give_powerup(player)`, because it means giving the pickup's "contained" powerup to the player

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
estebanfer: I still would prefer some comment so the people that see it in the API docs know better what it means without having to try it themselves, but I leave it up to you, resolve the convo or add some comment

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
estebanfer: `generate_landing_effects`, maybe `on_landing` if it's not only effects perhaps?

src/game_api/entities_items.hpp
```diff
@@ -91,12 +99,22 @@ class VladsCape : public Cape
 
 class Purchasable : public Movable
 {
-    virtual void buy(Entity* who) = 0;
+  public:
+    /// Be aware that for pickable items like weapons, the item will be automatically picked up by the player (unless he already holds something), the ownership transfer is a little bit slow, so it might trigger the shop keeper if you're not inside the shop.
+    /// does not consume money
+    virtual void acquire(Entity* who) = 0;
```
Mr-Auto, line 105: Btw. no objections to this name? I didn't like the "buy" name since it does not deal with the money and the ownership is janky as for the comment

src/game_api/entities_items.hpp
estebanfer: I guess I'll mark this as resolved now

src/game_api/entities_items.hpp
estebanfer: ah, I see, seems to be the function that gets called after the item is bought, to make the player to grab the item if possible (grab item, equip backitem), the comment for me made it seem like its a side effect of the function, rather than it's main thing.
And for the name I'm not sure, I guess it's fine as long as the doc comments is good enough? I think it's a little ambiguous but I'm not gonna be perfectionist, it could be fine as that if you also think so, and we don't come up with any better name

estebanfer: Testing a bit more, for powerup pickupables, calling it directly seems to just grant the powerup to the entity, just like being bought for free, same with backitems. But for items like guns, it just makes the player to grab the entity without doing anything else

docs/src/includes/_types.md
```diff
@@ -4224,37 +4239,46 @@ float | [offsetx](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=offse
...
 bool | [overlaps_with(Entity other)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=overlaps_with) | 
 [TEXTURE](#TEXTURE) | [get_texture()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=get_texture) | 
 bool | [set_texture(TEXTURE texture_id)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_texture) | Changes the entity texture, check the [textures.txt](game_data/textures.txt) for available vanilla textures or use [define_texture](#define_texture) to make custom one
-nil | [set_draw_depth(int draw_depth)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_draw_depth) | 
-nil | [set_enable_turning(bool enabled)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_enable_turning) | 
-nil | [liberate_from_shop()](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=liberate_from_shop) | 
+nil | [set_draw_depth(int draw_depth, int b3f)](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=set_draw_depth) |
```
estebanfer, line 4248: I forgot to say, the `set_draw_depth` function now requires an unknown b3f param, I think it would be fine, if anyone could have any idea what's b3f when using it. I know the old function is still usable, but it's NoDoc.
I think the best option is to either say what is b3f if you know, make b3f optional or else make the other function visible

src/game_api/entities_items.hpp
Mr-Auto: If i remember from testing, it removed the gun from the shop items.

docs/src/includes/_types.md
Mr-Auto: Well, the param is optional, just not shown as optional in the doc, i can try change that

I have no idea that's the purpose of it, game does something with it when going from one layer to another. But also there is default value in EntityDB, so no clue at all

estebanfer: > just not shown as optional in the doc, i can try change that
Yeah, that was my issue, the `spel2.lua` declarations will now complain every time you skip the parameter (as it already does with other things since some time, I should fix them also later). Also new people using the API would think it's required it the documentation says so

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
Mr-Auto: `generate_floor_landing_effects` ? fell like the floor is important here, since it does not work on stomping enemies, which can also be considered "landing on something". Or maybe just explain it in the comment since it's kind of long?

I still need to add comment that it does not generate the damage sounds or floor sounds when you fall while being stunned etc. (thrown by something for example)

src/game_api/entities_items.hpp
Mr-Auto: ok, so ... I think it's still valid function to detect purchase. Thou needs better comment and maybe better name?

docs/src/includes/_types.md
Mr-Auto: Changed it, the comment should show up after #399 is merged

src/game_api/script/usertypes/vtables_lua.cpp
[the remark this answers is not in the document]
estebanfer: Either the long name or comment would work. Landing alone may not be clear enough if stomping counts, but it's probably enough to make someone doubt if it does and check the comment

src/game_api/entities_items.hpp
estebanfer: if you don't wanna include the purchase action in it: `equip`, `try_equip`, `try_auto_equip`

Mr-Auto: the `on_` are usually problematic, as the hooks for me look stupid as `set_pre_on_something`

src/game_api/vtable_hook.hpp
Mr-Auto: Reverted this for now, it's kind of nice to have parameters that use reference and not pointer, to show that game does not do null check for it, but i guess it doesn't matter much since it is not shown in the API doc

src/game_api/sound_manager.hpp
estebanfer: Honestly, I can't come up with any name ideas, since I'm not completely sure what's the difference between those clases :cat2:.
Is the current SoundMeta for loop-able long sounds? Having a bool for `play_ending_secuence` makes it sound like it's repeatable and can be ended by game logic, and also the `start_over`

estebanfer: I guess you could leave it for another PR if you wanted

Mr-Auto: The `start_over` is more of a name i give it to make sense in the API rather then what actual functionality it holds.
The class name doesn't need to reflect much. In the source code it's probably just base class like "Sound".

That base class is used for stuff like: play and forget. You can just play sfx and not worry about it as the internals of it will deal with cleaning after it finishes etc.
Then `SoundMeda` is for sounds you want you either reuse or need to maintain (like keep updating the position etc.), it also used when you need to know if previous sound is still playing. Pretty sure you need to call the destroy function on it yourself

At the end is `BackgroundSound` which is probably only implemented for the main "music" in game_manager

There might be more that we don't know of, or that don't differ by the class members (only virtual overrides)

estebanfer: some rough ideas, `ContinuousSound`, `LoopableSound`? I suppose some examples could be the UFO laser shoot, lamassu laser prepare sound or Lamahu tentacles.
Do you have any ideas for the name yourself?

src/game_api/script/usertypes/vtables_lua.cpp
```diff
@@ -194,7 +194,7 @@ void register_usertypes(sol::state& lua)
         Entity,
         CallbackType::Entity,
         EntityVTable,
-        VTableEntry<"acquire", 93, MemFun<&Purchasable::acquire>>>;
+        VTableEntry<"acquire", 93, MemFun<&Purchasable::equip>>>;
```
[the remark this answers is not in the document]
Mr-Auto: yes xd

src/game_api/sound_manager.hpp
Mr-Auto: SFX -> Sound -> BackgroundSound  ?

estebanfer: Sounds pretty good
It could also be named SoundEffect and Sound changed into something like DynamicSound/ActiveSound to give contrast with the base class I guess
Also, is this gonna continue on this PR or should I approve this after testing everything works?

Mr-Auto: > Also, is this gonna continue on this PR or should I approve this after testing everything works?

We can move this conversation to discord and be done with this PR i guess

src/game_api/entities_items.hpp
```diff
@@ -739,7 +765,8 @@ class RollingItem : public Purchasable
   public:
     float roll_speed; // only positive numbers
 
-    virtual void on_purchase(Entity* who, int32_t unknown) = 0; // give you the powerup if you buy it
+    /// Skip this function for item to be unpickable
+    virtual void give_powerup(Entity* who, bool) = 0;
```
estebanfer, line 769: Well uh, now I found an issue here, skipping it doesn't seem to make pickups unpickable, and the boolean parameter seems to be for playing the sound or not (calling it with true plays the sound, with false it doesn't play). Probably named something like `play_sound` or `play_sfx`

estebanfer: Skipping it does prevent anything from happening when using `E:equip(P)` though

src/game_api/sound_manager.hpp
estebanfer: Alright

src/game_api/entities_items.hpp
Mr-Auto: fixed

estebanfer: no name given for the bool parameter?

Mr-Auto: i forgor

estebanfer (approved):
Approved! (finally)

PR #402, opened 2024-12-26
Level api lua additions

estebanfer:


src/game_api/level_api.hpp
```diff
@@ -538,14 +560,35 @@ struct LevelGenSystem
             LevelGenRooms* rooms_backlayer;
```
Mr-Auto, line 560: Can you change the `rooms_frontlayer` to `rooms_front_layer` to align with the renames below?

PR #407, opened 2024-12-30
State refactor

Mr-Auto:
- [x] implement struct for the custom allocated heap
- [x] move functions from `State` to `StateMemory` or standalone
- [x] find better place for all the hooks initialization
- [x] get rid of the `State` struct


src/game_api/thread_utils.hpp
```diff
@@ -1,57 +1,132 @@
...
-class OnHeapPointer
+struct HeapBase
 {
-    int64_t ptr_;
+    static HeapBase get(uint8_t slot);
+    static HeapBase get_main();
+    // can be NULL
+    static HeapBase get_local();
+    // fallback to main if can't get local
+    static HeapBase get_local_safe()
```
Mr-Auto, line 19: maybe instead of having `get_local` and `get_local_safe` i should flip them, so `get_local_unsafe` and `get_local` ?

src/game_api/thread_utils.hpp
Mr-Auto, line 1: The comparison looks bad, this is basically re-write. Was also thinking about changing the file name to just `heap_base`

src/game_api/thread_utils.hpp
Mr-Auto: or maybe even go minimalist approach, 'get` (which will be the local safe) and 'get_main`

estebanfer: I'm pretty sure the issue with local state being null is the way overlunky injects and runs some code in the thread that's injecting to it. If we make it to run code only on the spelunky thread, it shouldn't be null. iirc I got something to work in the third online changes PR, which isn't even open yet

Mr-Auto: for now i will just build in nullpointer check and make `get` and `get_main`

estebanfer: I think this is it
https://github.com/estebanfer/overlunky/commit/fc8b2c190a7558066448fefa89451a9c042efdc0#diff-20c19a6e71839b8083a6988dfb4bfb30a6598122ad4cffb83de6ffd418a176e7

Mr-Auto: ok cool, you want me to move that here?
but also, i needed to add this cuz it was crashing (don't remember if it was only PL maybe?)
https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/script/lua_vm.cpp#L147-L152

estebanfer: > ok cool, you want me to move that here?

if you want, iirc I just made that change quick and without thinking much, and it worked to make the local state valid at overlunky start, and I never got any issues elsewhere. I'm not sure if how it's coded it's really the best way to do it, but I guess it should be something like that

estebanfer: > but also, i needed to add this cuz it was crashing (don't remember if it was only PL maybe?)

hmm, I never got found any nullptr return using the local ptr with PL, not that I remember

src/game_api/thread_utils.hpp
estebanfer: Would make sense with that name too

src/game_api/illumination.cpp
```diff
@@ -18,6 +18,7 @@ Illumination* create_illumination(Vec2 pos, Color col, LIGHT_TYPE type, float si
 
         typedef Illumination* create_illumination_func(custom_vector<Illumination*>*, Vec2*, Color, LIGHT_TYPE, float, uint8_t light_flags, int32_t uid, uint8_t layer);
         static create_illumination_func* cif = (create_illumination_func*)(offset);
+        // enum_to_layer here does not use offset which you could argue should be used, since this function is comparable with spawn type function
```
Mr-Auto, line 21: Should i change this, or you recon this is too long like this and potentially breaking change for some mods is not worth it?

src/game_api/thread_utils.hpp
```diff
@@ -1,57 +1,120 @@
...
+    LiquidPhysics* liquid_physics() const noexcept
+    {
+        if (is_null())
+            return nullptr;
 
-    T* operator->() const
+        return reinterpret_cast<LiquidPhysics*>(ptr + GAME_OFFSET::LIQUID_ENGINE);
+    }
+
+    void copy_to(HeapBase other) const;
```
Mr-Auto, line 65: Was also thinking about making static function with two arguments. Should i or is this clear enough?

SereneRuby12: any of them seem clear enough to me

SereneRuby12 (changes requested):
Finally reviewing this. There's a memory leak that needs fixing and some small things, everything else is good!

src/game_api/entity.cpp
```diff
@@ -63,7 +64,7 @@ void Entity::set_layer(LAYER layer_to)
 
 void Entity::apply_layer()
 {
-    auto ptr_to = State::get().ptr()->layers[layer];
+    auto ptr_to = HeapBase::get().state()->layer(layer);
```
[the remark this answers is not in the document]
Mr-Auto: Since we expose the `layer` member without any restrictions,  i wanted to be extra safe and use safe version to get the layer ptr in case someone sets the layer to something other then 0/1, or the function `:remove()` was called on that entity.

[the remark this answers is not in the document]
Mr-Auto: Might be more reasonable to just fail if the layer is not 0 or 1, but didn't want to introduce too many unnecessary changes in this pr

src/game_api/script/lua_vm.cpp
```diff
@@ -1213,10 +1204,12 @@ end
     lua["lock_door_at"] = lock_door_at;
     /// Try to unlock the exit at coordinates
     lua["unlock_door_at"] = unlock_door_at;
-    /// Get the current frame count since the game was started. You can use this to make some timers yourself, the engine runs at 60fps. This counter is paused if you block PRE_UPDATE from running, and also doesn't increment during some loading screens, even though state update still runs.
-    lua["get_frame"] = get_frame_count;
+    /// Get the current frame count since the game was started*. You can use this to make some timers yourself, the engine runs at 60fps. This counter is paused if the pause is set with flags PAUSE.FADE or PAUSE.ANKH.
```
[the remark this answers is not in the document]
Mr-Auto: It's paused with flag `PAUSE.FADE` or flag 8 (no enum for that)
can also be reverted with the multiplayer rollback 

Either forgot to explain or didn't know how

src/game_api/heap_base.cpp
```diff
@@ -0,0 +1,159 @@
...
+    if (slot >= MAX_SAVE_SLOTS)
+        return NULL;
+    static HeapBase* save_slots = reinterpret_cast<HeapBase*>(get_address("save_states"));
+    return *(save_slots + slot);
+}
+
+void HeapClone(HeapBase heap_to, uint64_t heap_container_from)
+{
+    auto heap_from = memory_read<uint64_t>(heap_container_from + 0x88);
+    HeapBase heap_base_from = reinterpret_cast<HeapBase&>(heap_from);
```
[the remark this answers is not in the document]
Mr-Auto: not my code https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/state.cpp#L703
I can change it so it just reads the address from `heap_container_from` and creates local HeapBase instead, would that be fine?

src/game_api/savestate.hpp
```diff
@@ -1,19 +1,52 @@
...
+        save();
+    }
+    /// NoDoc
+    SaveState(uint8_t index)
+        : base(HeapBase::get(index)){};
     ~SaveState()
     {
         clear();
     }
+    /// Get the pre-allocated by the game save slot 1-4. Call as `SaveState.get(slot)`
```
Mr-Auto, line 26: Should i change it so it's `SaveState:get(slot)`?
as i don't think we have any functions called with just the `.` dot

src/game_api/entity.cpp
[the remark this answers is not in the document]
Mr-Auto: As for the others in this file
`set_layer` is safe, since there is check for layer 0 and 1, and enum_to_layer always returns 0 or 1

`remove` i guess could get better check

src/game_api/heap_base.cpp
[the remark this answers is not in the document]
SereneRuby12: I mean its a bit different casting a pointer type to another pointer type, to casting a variable to be a different type (both non-pointer).
I guess it's not thaaat different but it's way more confusing to understand how that cast even works.

> I can change it so it just reads the address from heap_container_from and creates local HeapBase instead, would that be fine?

Oh yeah that seems fine if I get what you're saying correctly

src/game_api/savestate.hpp
SereneRuby12: ImGuiIo does use the dot function call, though idk if it's generated well in `spel2.lua`

src/game_api/heap_base.cpp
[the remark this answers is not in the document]
SereneRuby12: Though I'm a bit concerned if something happens and we add another member variable to HeapBase (like how you added a new one to SaveState), and this breaks. Maybe it's just me getting weirded by casting a primitive type to a struct, idk

src/game_api/entity.cpp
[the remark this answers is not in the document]
SereneRuby12: ah makes sense, and `remove` would need extra code that's not really needed in this PR

src/game_api/heap_base.cpp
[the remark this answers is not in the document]
Mr-Auto: > Though I'm a bit concerned if something happens and we add another member variable to HeapBase (like how you added a new one to SaveState), and this breaks. 
Yeah, that's why i want to read/copy the address and construct local `HeapBase` from it instead of casting the unknown memory to `HeapBase`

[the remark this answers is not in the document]
SereneRuby12: > Yeah, that's why i want to read/copy the address and construct local HeapBase from it instead of casting the unknown memory to HeapBase

Alright that's good

src/game_api/script/lua_vm.cpp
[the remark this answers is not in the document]
Mr-Auto: Rewritten the comment a little

src/game_api/savestate.hpp
Mr-Auto: spel2.lua generates `function SaveState:get(save_slot) end` so i guess i should change this
For `modifierdown` it generates the standard format, so probably suggests `:`, dunno if there is a way to suggest a dot, i don't use it myself

SereneRuby12: it's an issue for another time, we already have things in the API with the same problem, so I think it would be better to solve this in a specific PR for this

src/game_api/script/lua_vm.cpp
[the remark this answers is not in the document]
SereneRuby12: seems fine

src/game_api/savestate.hpp
Mr-Auto: k

SereneRuby12:
hmm, checking some behavior, it seems weird to me that `load_state(1)` checks for `state->screen` but `SaveState.get(1):load()` doesn't

Mr-Auto:
> hmm, checking some behavior, it seems weird to me that `load_state(1)` checks for `state->screen` but `SaveState.get(1):load()` doesn't

I literally asked about this on discord xd

Mr-Auto:
" being able to load **invalid** the game savestate thru SaveState class"
forgot important word in last commit

SereneRuby12 (approved):
All good :D

PR #410, opened 2025-01-01
Improve `illumination.lua`

Mr-Auto:


docs/src/includes/_globals.md
```diff
@@ -2273,6 +2273,7 @@ Current mouse cursor position in screen coordinates.
 #### [Illumination](#Illumination) create_illumination([Color](#Color) color, float size, int uid)
 
 Creates a new [Illumination](#Illumination). Don't forget to continuously call [refresh_illumination](#refresh_illumination), otherwise your light emitter fades out! Check out the [illumination.lua](https://github.com/spelunky-fyi/overlunky/blob/main/examples/illumination.lua) script for an example.
+Warning: this is only valid for current level!
```
Mr-Auto, line 2276: Should i also add stuff like "save the `Illumination` in `state.user_data` for it to work online" ?

Mr-Auto: On one hand seams logical, on the other, this should be rule for everything, not just the illumination

estebanfer: does it even work online correctly yet? isn't it a pointer?

Mr-Auto: hmm... actually you're right, i was only thinking about rollback from it being created to when it wasn't. But yeah, it won't work when the heap is copied to different one.
Unless we can adjust the pointer in user_data just like the game adjust it's pointers

Mr-Auto: or, i can make some weird solution for the meta table so it can use OnHeapPointer

estebanfer: both solutions would work I guess, just that I have no idea how to know if a lua object is a pointer for the first idea

Mr-Auto: > both solutions would work I guess, just that I have no idea how to know if a lua object is a pointer for the first idea

```cpp
[](sol::object obj)
    {
        if (obj.valid() && obj.get_type() == sol::type::userdata)
        {
            DEBUG("{:x}", *(size_t*)obj.pointer());
        }
    };
```

Mr-Auto: Didn't test much, but i think all userdata are stored as pointers to an actual data, even stuff created on lua stack like when using `Color:new()`

estebanfer: if it works, and is documented that it does that, it seems a fine possible solution

PR #411, opened 2025-01-09
Even more refactor

Mr-Auto:
Looking for way to reduce `lua_vm.cpp` size. Maybe also `rpc.cpp` and other things that feel usefull

Continuation of #407


src/injected/ui_util.cpp
```diff
@@ -472,13 +468,13 @@ void UI::update_floor_at(float x, float y, LAYER l)
     if (uid == -1)
         return;
     auto ent = get_entity_ptr(uid);
-    if ((ent->type->search_flags & 0x100) == 0 || !test_flag(ent->flags, 3))
+    if (!(ent->type->search_flags & ENTITY_MASK::FLOOR) || !test_flag(ent->flags, 3))
```
Mr-Auto, line 471: Is this fine, or should i be using `!= ENTITY_MASK::X` wherever possible?

src/game_api/entity.cpp
```diff
@@ -321,27 +321,28 @@ void Movable::set_position(float to_x, float to_y)
...
 {
-    auto acutal_mask = [](uint32_t m) -> uint32_t // for the MASK.ANY
-    { return m == 0 ? 0xFFFF : m; };
+    auto actual_mask = [](ENTITY_MASK m) -> ENTITY_MASK // for the MASK.ANY
+    { return m == ENTITY_MASK::ANY ? (ENTITY_MASK)0xFFFF : m; };
 
     if (rec_mode == RECURSIVE_MODE::EXCLUSIVE)
     {
-        if (mask.has_value() && (acutal_mask(mask.value()) & ent->type->search_flags) != 0)
+        if (mask.has_value() && !!(actual_mask(mask.value()) & ent->type->search_flags))
```
Mr-Auto, line 337: This approach `!!` is very debatable online, so I'm just highlighting it beforehand.
The only alternative I think is using function for it

Mr-Auto:
With this, files are more or less around 1k lines
exceptions:
- level_api - mostly due to the community chances and tilecodes adding around 500 lines
- search - for obvious reasons
- string_hashes - for obvious reasons
- lua_backend - due to all the events, i think @SereneRuby12 is working template for this?
- lua_vm - is still almost 2k, thou 500 lines just for the `ON` enum and it's comments (without all the comments the file is just 1400 lines now)
- entity_monsters_lua - it's just so many sub-classes to expose, feels weird to split it
- gui_lua - about 200 lines of Deprecated functions, plus a lot of stuff in general
- level_lua - mostly CustomTheme stuff - should probably move to separate file
- info_dump - don't care, thou mostly listing values from one enum
- ui.cpp - **begs for refactor**
- don't really care about python scripts or other files

Mr-Auto:
Finished self review of this, made sure that nothing was lost in the doc, just moved around
90% of this PR is just moving stuff around

src/injected/ui_util.cpp
SereneRuby12: I think it's fine as long as it works, and it's still readable

src/game_api/entity.cpp
SereneRuby12: I don't think we have any rules on this, as long as it's readable and works well, I think it's fine

PR #415, opened 2025-02-07
Reverse engeneer more of the `Renderer` struct and `Texture`

Mr-Auto:


src/game_api/texture.cpp
```diff
@@ -274,34 +276,40 @@ void reload_texture(const char** texture_name)
 
     auto renderer_ptr = render.renderer();
     auto load_texture = *vtable_find<LoadTextureFunT*>(renderer_ptr, c_LoadTextureVirtualIndex);
+
+    if (texture_name->dx_resource)
+    {
+        texture_name->dx_resource->release(); // should be safe since apparently the objects have reference counter, essentially working like shared pointer
+        game_free(texture_name->dx_resource);
+    }
     load_texture(renderer_ptr, texture_name);
```
Mr-Auto, line 285: This needs more testing, hard to tell if this is correct since game never reloads textures

But the `load_texture` call does create new `dx_resource` object and just overrides the old one without even checking (normally `dx_resource` is set to the placeholder texture when `Resource` is created, and then this function makes proper object for the texture and overrides the placeholder)

src/info_dump/main.cpp
```diff
@@ -974,7 +974,7 @@ void get_vtables()
         {
             if (it.id)
             {
-                const size_t record = (size_t) * (offsets.begin() + idx);
+                const size_t record = (size_t)*(offsets.begin() + idx);
```
Mr-Auto, line 977: This it due to some Visual Studio Community update. But actually the current clang format feels wrong for this, makes it tool like multiplication.

And according to our own clang format rules the de-refrence should not have a space between the pointer and `*`, Same for the C-style casting

src/game_api/texture.cpp
```diff
@@ -274,34 +276,40 @@ void reload_texture(const char** texture_name)
...
 }
 
 void set_heart_color_from_texture(TEXTURE vanilla_id, TEXTURE custom_id)
 {
     uint32_t char_index = (uint32_t)vanilla_id - 285;
-    if (char_index >= 0 && char_index <= 19)
+    if (char_index > 19)
+        return;
```
Mr-Auto, line 292: Only inverted this if check, but the indentation change shows as some bigger difference

PR #416, opened 2025-02-08
Godmode fix

### LINKED ISSUE #381: Godmode affects unrelated player movement
Godmode affects player movement mechanics that are unrelated to taking damage or debuffs. I've found two cases where this happens, both caused by godmode forcing `lock_input_timer` to 0.

- Godmode lets you start moving after using a layer door one frame earlier than you would be able to otherwise. `lock_input_timer` is normally set to 1 for the duration of a layer door usage and takes one extra frame to expire after finishing the door usage.
- Godmode clears the 10 frames of locked input after firing a weapon with recoil, which affects the behavior of things like shotgun jumps if the player is holding inputs during that recoil period.

What was the intended purpose of setting `lock_input_timer` to 0? Was it to clear the input lockout that happens if you touch an enemy while airborne? If you can't get hit anyways in godmode, then that would only matter if a user took a hit without godmode and then tried to switch to godmode immediately to clear the input lockout.

Mr-Auto:
- don't force `lock_input_timer` to 0 every frame
- fix the floating in the air when fallen into item spikes (movable spikes on hundun platform)

Fixes #381
`lock_input_timer` is also used by the tentacles at the bottom of tiamat and abzu to lock you out of controls before dropping into the abyss. Thou i don't think it's that big of a deal

### DIFF
--- a/src/injected/ui.cpp
+++ b/src/injected/ui.cpp
@@ -1943,10 +1943,12 @@ void force_cheats()
             ent->poison_tick_timer = -1;
             ent->onfire_effect_timer = 0;
             ent->wet_effect_timer = 0;
-            ent->lock_input_timer = 0;
             ent->set_cursed(false, false);
             ent->more_flags &= ~(1U << 16);
             UI::destroy_entity_item_type(ent, ink);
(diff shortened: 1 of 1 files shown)


SereneRuby12 (approved):
All good!

PR #420, opened 2025-03-07
temp clang version mismatch fix

Mr-Auto:


SereneRuby12:
do we know if this is safe to do? or if there could be some issues due to using an older version?

Mr-Auto:
"Mixing an older STL with a newer compiler is untested and unsupported, but will typically work"
I assume the same for the other way around
https://devblogs.microsoft.com/cppblog/c17-progress-in-vs-2017-15-5-and-15-6/
It is temporary fix after all. 

Everything seam to work when i tested it.
Just to be sure we could post the latest whip as standalone release to fall-back to if this ever breaks anything

SereneRuby12:
If included with a way to fall-back, I think this would be fine

Mr-Auto:
Sure, if latest whip will break, will just delete it so the backup one will be the latest

PR #422, opened 2025-03-22
Fix entity casting as function return

Mr-Auto:


src/game_api/script/usertypes/entity_lua.cpp
```diff
@@ -242,8 +230,12 @@ void register_usertypes(sol::state& lua)
     entity_type["offsety"] = &Entity::offsety;
     entity_type["rendering_info"] = &Entity::rendering_info;
     entity_type["user_data"] = std::move(user_data);
-    entity_type["topmost"] = topmost;
-    entity_type["topmost_mount"] = topmost_mount;
+    /// Returns the top entity in a chain (overlay)
+    entity_type["topmost"] = [&lua](Entity& entity) // -> Entity
```
Mr-Auto, line 234: I assume we would rather like to pull comments from the actual function instead of copying them here?
or is whatever?

PR #424, opened 2025-03-25
Exposes a bunch of FMOD Studio stuff to Lua

doofus9341:
This adds the ability to load FMOD Studio bank files, get FMOD Event Descriptions, and create and control FMOD Studio Events in Lua. The ability to create custom FMOD Studio Events is extremely powerful and allows you do a bunch of audio stuff that was previously difficult like seamless custom music, dynamic music, custom level ambience, shop music, and audio effects.

I have created a blank [FMOD Studio template project](https://github.com/doofus9341/Spel2-FMOD-Template) that currently only builds a dummy master bank in order to make it easy to create new FMOD Banks and Events. The project has been manually edited to contain matching group and return buses with the correct GUIDs. This allows custom FMOD Events to be routed into Spelunky 2's FMOD Studio Mixer using sends or by placing the Event in the desired group bus. A benefit of this is that it allows the game to automatically apply various effects to audio to match vanilla behavior like ducking music and SFX when paused, in level transition screens, or gradually fading music out upon death.

New usertypes:

 - `CustomBank` is a handle to a loaded FMOD bank file. Can be used to unload the bank file, or control the loading state of non-streamed sample data in the bank.
 - `CustomEventDescription` is a handle to an FMOD Event Description. Can be used to get parameter IDs, create instances of that event, or control the loading state of sample data for that event.
 - `CustomEventInstance` is a handle to an FMOD Event Instance. Provides a lot of control over event instances; allows you to set and get parameters, the instances timeline position, volume, and pitch.
 - `ParameterDescription` is a description of an FMOD Event parameter, can be used to retrieve a parameter ID and view other information about the parameter such as flags, or the minimum, maximum, and default value.
 - `ParameterId` is an FMOD Event parameter identifier. This usertype does not have functions but can be used to quickly set or get a parameter for an Event Instance without FMOD having to do slower name to ID lookups.
 - `FMODguidMap` has a single purpose currently. It allows you to create a `CustomEventDescription` by looking up the GUID from a human readable event path. Normally FMOD looks up the GUIDs of events using a string table stored in the strings bank which is bundled with the master bank. Since the strings bank can't be rebuilt to include new event paths and GUIDs. It parses the GUIDs.txt file exported from an FMOD Studio project which includes the path and GUID of Events and Snapshots.

New functions:

 - `load_bank` takes a file path and allows you to load FMOD Studio Bank files from the disk. Returns a `CustomBank`.
 - `get_bank` takes a file path and allows you to get an existing `CustomBank` if it has already been loaded.
 - `get_event_by_id` takes an FMOD GUID string and retrieves a `CustomEventDescription` assuming it has been loaded by FMOD Studio.
 - `create_fmod_guid_map` takes a file path and parses the GUIDs.txt file exported from an FMOD Studio project. It returns a `FMODguidMap` that can be used to lookup the GUIDs of Events and Snapshots and create a `CustomEventDescription`.

Resolved issues:
- FMOD Event Instances are not automatically released, meaning that if you do not call `release()` on a `CustomEventInstance` the Event Instance will never be marked for release and destroyed. Event Instances are also not automatically stopped meaning that looping event instances won't be destroyed even if they have been marked for release unless they are manually stopped. Unloading the bank file the Event Instance was loaded from will destroy all Event Instances and Event Descriptions from that bank as expected.
   - Resolved in 54929d9a2cd14059e84eb4a7a2e16b9994502762
- Bank reference counting is janky. It mostly works, but I may wind up changing it or removing it completely.
   - Removed in 4cf2a5cb948e1a41877ee2a0f784a29e74c06cd0

Known issues:
- FMOD Banks are not automatically unloaded when a script unloads, so script mods may have Event Instances, Event Descriptions, and Banks persist after the script has been unloaded.
- Calling FMOD functions after SoundManager has been destroyed is not safe; e.g. calling functions while exiting the game in an ON.SCRIPT_DISABLE callback.
   - I have resolved this issue on the [soundmanager-shared-ptr](https://github.com/doofus9341/overlunky/tree/soundmanager-shared-ptr) branch of my fork, however I am unsure whether using shared pointers for this is the right call. If using shared pointers is acceptable, I can open a separate pull request to resolve this issue.


docs/src/includes/_types.md
```diff
@@ -876,6 +959,27 @@ int | [random(int min, int max)](https://github.com/spelunky-fyi/overlunky/searc
...
+ | [type](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=type) | 
+ | [flags](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=flags) | 
+
+### ParameterId
+
+
+Type | Name | Description
+---- | ---- | -----------
+ | [data1](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=data1) | 
+ | [data2](https://github.com/spelunky-fyi/overlunky/search?l=Lua&q=data2) |
```
Mr-Auto: Missing type names

Mr-Auto (changes requested):

docs/src/includes/_types.md
doofus9341: Fixed in 438ffb3d659292b8067ee450c4bb25cb046d634f.

PR #426, opened 2025-04-29
Fix current callback being overwritten if another one starts before the first one finished execution

Mr-Auto:
Daft because i didn't test if there are possible situations when the calls to `set_current_callback` and `clear_current_callback` are uneven

### DIFF
--- a/src/game_api/script/lua_backend.hpp
+++ b/src/game_api/script/lua_backend.hpp
@@ -467,7 +467,7 @@ class LuaBackend
         current_cb.emplace(aux_id, id, type);
 
         return OnScopeExit([&]()
-                           { if (!current_cb.empty()) current_cb.pop(); });
+                           { if (!current_cb.empty()) current_cb.pop(); else DEBUG("Trying to pop empty current_callback stack (aux: {} id: {} type: {})\n", aux_id, id, (int)type); });
     }
 
     void set_error(std::string err);
--- a/src/game_api/script/lua_backend.hpp
+++ b/src/game_api/script/lua_backend.hpp
@@ -5,6 +5,7 @@
 #include <cstdint>       // for uint32_t, uint16_t, uint8_t, int32_t
 #include <deque>         // for deque
 #include <filesystem>    // for path
+#include <forward_list>  // for forward_list
 #include <functional>    // for equal_to, function, less
 #include <imgui.h>       // for ImDrawList (ptr only), ImVec4
 #include <locale>        // for num_get, num_put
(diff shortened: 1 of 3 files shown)


SereneRuby12:
I see, probably it would be better to use std::stack like `g_callingBackend` does here https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/script/lua_backend.cpp#L1760C51-L1760C67

SereneRuby12:
Seems that those can be uneven at `LuaBackend::pre_get_feat`, `LuaBackend::pre_get_feat`, `LuaBackend::pre_set_feat` due to the functions being able to return early

Mr-Auto:
> I see, probably it would be better to use std::stack like `g_callingBackend` does here https://github.com/spelunky-fyi/overlunky/blob/main/src/game_api/script/lua_backend.cpp#L1760C51-L1760C67

I can make it stack, it makes almost no difference

> I guess the util `ON_SCOPE_EXIT` that's used to pop the calling backend from the stack could be useful for this too

I was thinking about something like that to enure both gets called

src/game_api/script/lua_backend.hpp
```diff
@@ -454,9 +455,20 @@ class LuaBackend
...
+            return {0, 0, CallbackType::None};
+
+        return current_cb.top();
+    }
+    [[nodiscard]] auto set_current_callback(int32_t aux_id, int32_t id, CallbackType type)
+    {
+        current_cb.emplace(aux_id, id, type);
+
+        return OnScopeExit([&]()
+                           { if (!current_cb.empty()) current_cb.pop(); });
```
Mr-Auto, line 470: The `if` here is probably unnecessary, but the "undefined behavior" always scares me

SereneRuby12 (approved):
Everything seems fine

src/game_api/script/lua_backend.hpp
```diff
@@ -454,9 +455,20 @@ class LuaBackend
...
+            return {0, 0, CallbackType::None};
+
+        return current_cb.top();
+    }
+    [[nodiscard]] auto set_current_callback(int32_t aux_id, int32_t id, CallbackType type)
+    {
+        current_cb.emplace(aux_id, id, type);
+
+        return OnScopeExit([&]()
+                           { if (!current_cb.empty()) current_cb.pop(); else DEBUG("Trying to pop empty current_callback stack (aux: {} id: {} type: {})\n", aux_id, id, (int)type); });
```
SereneRuby12, line 470: Uhm I'm a bit late but wouldn't the stack variables need to be captured by value rather than by reference?

Mr-Auto: uh oh, will fix on main

